PR #8, opened 2018-11-15
Forman 7 grid tool

### LINKED ISSUE #7: Allow aligning regional cubes on a fixed Earth grid
Develop a CLI that allows computing suitable Xcube spatial resolutions and allows for adjusting a given bounding box so that it snaps into fixed Earth multi-resolution pyramid grid.

forman:
In this PR, I re-coded `xcube-grid` CLI from scratch! Note, it is **important that we agree on the logic implemented here**, because we will need to apply it to all the sub-regions in the DCS4COP and CyanoAlert projects. 

Could you please check it out and test it. In the `README.md` is a related section that explains the logic also provides an example use case. See also section in `docs/XCUBE-SPEC.md`.

Closes #7.


forman:
Ja, der implementiereten Logik folgend, ist das sogar richtig, aber leider nicht zielführend.

forman:
Fixed

PR #24, opened 2019-02-04
New API and CLI

forman:


xcube/genl2c/default/inputprocessor.py
```diff
@@ -56,7 +56,7 @@ def name(self) -> str:
 
     @property
     def description(self) -> str:
-        return 'Single-scene NetCDF inputs in xcube standard format'
+        return 'Single-scene NetCDF/CF inputs in xcube standard format'
```
AliceBalfanz: Test needs to be adjusted to the change:
/xcube/test/genl2c/default/test_inputprocessor.py - line 18

Issue #13, opened 2018-12-19
expand default processor of xcube-genl2 to handle  0-360 lon 


AliceBalfanz:
The default processor only accepts netcdf standard -180 to 180 longitudes and does not normalize input of longitude 0-360.

forman:
See normalisation implementation  in Cate: https://github.com/CCI-Tools/cate/blob/master/cate/core/opimpl.py#L163

forman:
Gunnar just pointed me to this cool function which you can use too: http://xarray.pydata.org/en/stable/generated/xarray.DataArray.roll.html

forman:
@AliceBalfanz, for the current implementation, please note that

```
test/api/gen/default/test_gen.py::DefaultProcessTest::test_handle_360_lon
  D:\Projects\xcube\xcube\api\gen\default\iproc.py:183: FutureWarning: roll_coords will be set to False in the future. Explicitly set roll_coords to silence warning.
    dataset = dataset.roll(lon=lon_size_05)
```

Issue #9, opened 2018-11-28
Switch from argparse package to click package for xcube-genl2c and xcube-genl3


AliceBalfanz:
The arguments given by the command line are currently extracted by the argparse package. When passing negative values for an argument. The click packages allows to enable negative values without returning an error.

PR #29, opened 2019-02-11
Changing cli.gen from argparse to click

AliceBalfanz:
Changing cli.gen form argparse to click, implementation of issue #9  
Help section needs to be extended in future (issue #13)


test/cli/test_gen.py
```diff
@@ -1,46 +1,55 @@
...
+
+    def test_missing_args(self):
+        result = self.invoke_cli([])
         # exit_code==0 when run from PyCharm, exit_code==2 else
-        self.assertEqual(2, exit_code)
-        #self.assertTrue(exit_code in [0, 2])
+        self.assertEqual(1, result.exit_code)
 
     def test_main_with_illegal_size_option(self):
-        exit_code = main(['-s', '120,abc', 'input.nc'])
```
forman: For me, the name `exit_code` was clearer because it also reflected the role of the return value. In fact, the CLI will use this value as exit code in the shell.

xcube/api/gen/defaults.py
```diff
@@ -21,6 +21,6 @@
 
 DEFAULT_OUTPUT_DIR = '.'
 DEFAULT_OUTPUT_NAME = 'PROJ_WGS84_{INPUT_FILE}'
-DEFAULT_OUTPUT_WRITER = 'nc'
+DEFAULT_OUTPUT_FORMAT = 'zarr'
```
forman: Good!

xcube/cli/gen.py
```diff
@@ -19,80 +19,86 @@
...
-    traceback_mode = arg_obj.traceback_mode
-    append_mode = arg_obj.append_mode
-    dry_run = arg_obj.dry_run
+    traceback_mode = traceback
+    append_mode = append
+    dry_run = dry_run
 
     try:
-        config = get_config_dict(arg_obj, open)
+        config = get_config_dict(locals(), open)
```
forman: Why `locals()`?

xcube/cli/gen.py
```diff
@@ -110,41 +116,40 @@ def _handle_error(e, traceback_mode):
...
+#         return help_text
+#
+#     @classmethod
+#     def _format_dataset_ios(cls, dataset_ios):
+#         help_text = ''
+#         for ds_io in dataset_ios:
+#             fill1 = ' ' * (24 - len(ds_io.name))
+#             fill2 = ' ' * (10 - len(ds_io.ext))
+#             help_text += f'  {ds_io.name}{fill1}(*.{ds_io.ext}){fill2}{ds_io.description}\n'
+#         return help_text
```
forman: Please find a solution how to display this info. If there is absolutley no chance to customize click, then we need to introduce a new command, such as `info` with special options that will print out the desired information, e.g.

    $ xcube info --formats
    $ xcube info --iprocs

AliceBalfanz: This will be solved in a different branch as mentioned in the pull request text at the very top. It is also remarked in with a TODO in this branch.

xcube/cli/gen.py
AliceBalfanz: it is now solved by keeping the structure of the old code.

test/cli/test_gen.py
```diff
@@ -26,8 +26,8 @@ def test_missing_args(self):
         self.assertEqual(1, result.exit_code)
 
     def test_main_with_illegal_size_option(self):
-        result = self.invoke_cli(['-s', '120,abc', 'input.nc'])
-        self.assertEqual(1, result.exit_code)
+        exit_code = self.invoke_cli(['-s', '120,abc', 'input.nc'])
```
forman: I was not aware that `result`*has an* `exit_code`. Therefore "result" was fine, SORRY!

xcube/cli/gen.py
```diff
@@ -90,12 +90,17 @@ def generate_cube(input_files: str,
     Level-2C data cubes may be created in one go or successively in append mode, input by input.
     The input may be one or more input files or a pattern that may contain wildcards '?', '*', and '**'.
     """
+    try:
+        arg_obj = locals()
```
forman: This still doesn't explain why`locals()` is used here.

xcube/cli/gen.py
```diff
@@ -90,12 +90,17 @@ def generate_cube(input_files: str,
...
+    except SystemExit as e:
+        return int(str(e))
+
     traceback_mode = traceback
     append_mode = append
     dry_run = dry_run
 
     try:
-        config = get_config_dict(locals(), open)
+        config = get_config_dict(arg_obj, open)
```
forman: No difference, effectively.

xcube/cli/gen.py
```diff
@@ -90,12 +90,17 @@ def generate_cube(input_files: str,
     Level-2C data cubes may be created in one go or successively in append mode, input by input.
     The input may be one or more input files or a pattern that may contain wildcards '?', '*', and '**'.
     """
+    try:
+        arg_obj = locals()
+    except SystemExit as e:
```
forman: A call to ` locals()` does not raise any exception, afaik.

xcube/cli/gen.py
forman: I do not agree. This feature is mandatory and cannot be dropped in our code on the master branch.

xcube/cli/gen.py
forman: Please try understanding the old code first, before changing its structure. You may unintentionally change the logic. Passing `locals()` as arg_obj doesn't make sense, because locals does not have the expected attributes, please look into the implementation of `get_config_dict()`.

B.t.w. where is the new logic that accpets multiple `-c <config>` options?

xcube/cli/gen.py
```diff
@@ -111,40 +131,52 @@ def _handle_error(e, traceback_mode):
     return 2
 
 
-class GenL2CHelpFormatter(argparse.HelpFormatter):
+@click.command()
+@click.option('--format', '-f', metavar='OUTPUT_FORMAT', default=False, is_flag=True,
```
forman: if `is_flag=True` then `metavar` doesn't make sense.

xcube/cli/gen.py
```diff
@@ -111,40 +131,52 @@ def _handle_error(e, traceback_mode):
...
+              help='Additional help information about formats. '
+              f'The choices for the output format are: {output_writer_names}')
+@click.option('--proc', '-p', metavar='INPUT_PROCESSOR', default=False, is_flag=True,
```
forman: if `is_flag=True` then `metavar` doesn't make sense.

xcube/cli/gen.py
```diff
@@ -111,40 +131,52 @@ def _handle_error(e, traceback_mode):
...
+              help="Additional help information about input processors. "
+              f'The choices as input processor are: {input_processor_names}')
+def info(format: bool, proc: bool):
+    """
+    Displays additional information about format options or about input processors.
+    """
+    if format is True:
+        click.echo(format_help('format'))
+    if proc is True:
+        click.echo(format_help('proc'))
```
forman: So if I provide no option, then nothing is displayed?

PR #30, opened 2019-02-15
Alicja merge config files current version

AliceBalfanz:
adding possibility of using several configuration files, which then are merged into one and used for the xcube generation. The order of the configuration files matters, the first configuration file is allways updated by the others passed afterwards.


test/api/gen/test_config.py
```diff
@@ -26,20 +17,25 @@ def get_config_obj(config_file=None,
...
+                output_variables=output_variables,
+                output_resampling=output_resampling)
+
 
+TEMP_PATH_FOR_YAML = 'temp_test_data_for_xcube_tests'
 
 CONFIG_1_FILE = 'config_1.json'
+CONFIG_1_FILE_2 = 'config_1_2.json'
+CONFIG_1_FILE_LIST = [(os.path.join(os.path.expanduser('~'), TEMP_PATH_FOR_YAML, CONFIG_1_FILE)),
+                      (os.path.join(os.path.expanduser('~'), TEMP_PATH_FOR_YAML, CONFIG_1_FILE_2))]
```
forman: Never write in user'home!

xcube/api/gen/config.py
```diff
@@ -80,7 +79,7 @@ def get_config_dict(config_obj, open_function):
         except ValueError:
             output_size = None
         if output_size is None or len(output_size) != 2:
-            raise ValueError(f'invalid output_size {config_obj.output_size!r}')
+            raise ValueError(f'invalid output_size {config_obj.get("output_size")!r}')
```
forman: The value referred to in the error message is a not the object being tested in the `if` condition.

xcube/api/gen/config.py
```diff
@@ -89,7 +88,7 @@ def get_config_dict(config_obj, open_function):
         except ValueError:
             output_region = None
         if output_region is None or len(output_region) != 4:
-            raise ValueError(f'invalid output_region {config_obj.output_region!r}')
+            raise ValueError(f'invalid output_region {config_obj.get("output_region")!r}')
```
forman: The value referred to in the error message is a not the object being tested in the `if` condition.

xcube/api/gen/config.py
```diff
@@ -18,78 +18,82 @@
...
         try:
             output_region = list(map(lambda c: float(c), output_region.split(',')))
         except ValueError:
-            output_region = None
-        if output_region is None or len(output_region) != 4:
-            raise ValueError(f'invalid output_region {config_obj.output_region!r}')
+            raise ValueError(f'Invalid output region was given. Only floats are accepted. The given output region was: '
+                             f'{config_obj.get("output_region")!r}')
+        if len(output_region) != 4:
+            raise ValueError(f'The length of the output region is not 4.The given output region was: '
```
forman:
```suggestion
            raise ValueError(f'The length of the output region is not 4. The given output region was: '
```

xcube/api/gen/config.py
```diff
@@ -18,78 +18,82 @@
...
             output_size = list(map(lambda c: int(c), output_size.split(',')))
         except ValueError:
-            output_size = None
-        if output_size is None or len(output_size) != 2:
-            raise ValueError(f'invalid output_size {config_obj.output_size!r}')
+            raise ValueError(
+                f'Invalid output size was given. Only integers are accepted. The given output size was: '
+                f'{config_obj.get("output_size")!r}')
+        if len(output_size) != 2:
+            raise ValueError(f'The length of the output size is not 2. The given length for output size is: '
```
forman:
```suggestion
            raise ValueError(f'The output size must be given as pair <width>,<height>, but was: '
```

xcube/api/gen/config.py
```diff
@@ -18,78 +18,82 @@
...
+        for config_path in config_paths:
+            with open(config_path) as fp:
+                config_dicts.append(yaml.load(fp))
+        config = merge_config(*config_dicts)
 
-                config = yaml.load(stream)
-        except yaml.YAMLError as e:
-            raise ValueError(f'YAML in {config_file!r} is invalid: {e}') from e
-        except OSError as e:
-            raise ValueError(f'cannot load configuration from {config_file!r}: {e}') from e
```
forman: Why has the error handling be removed? By what has it been replaced?

xcube/api/gen/config.py
```diff
@@ -18,78 +18,82 @@
...
+    output_resampling = config_obj.get("output_resampling")
 
     if config_file is not None:
-        try:
-            with open_function(config_file) as stream:
+        config_paths = config_file
+        config_dicts = []
+        for config_path in config_paths:
+            with open(config_path) as fp:
+                config_dicts.append(yaml.load(fp))
```
forman:
```suggestion
            config_dict = yaml.load(fp)
        except yaml.YAMLError as e:
            raise ValueError(f'YAML in {config_file!r} is invalid: {e}') from e
        except OSError as e:
            raise ValueError(f'cannot load configuration from {config_file!r}: {e}') from e
        if not isinstance(config_dict, dict):
            raise ValueError(f'invalid configuration format in {config_file!r}: dictionary expected')
        config_dicts.append(config_dict)
```

PR #36, opened 2019-03-07
Convert vars to dims

dzelge:
When accepting the PR a new API function (```vars_to_dim```) will be introduced that converts the 
variables of  a data cube into a single dimension.


xcube/cli/cli.py
```diff
@@ -118,6 +118,42 @@ def dump(dataset, variable, encoding):
         print(text)
 
 
+# noinspection PyShadowingBuiltins
+@click.command(name="vars2dim")
+@click.argument('cube', metavar='<cube>')
+@click.option('--dim_name', '-d', metavar='<dimension name>', help="Name of the new dimension ['var']")
+@click.option('--output', '-o', metavar='<output>',
+              help="Output file.")
```
AliceBalfanz: It might be clearer to have 
`help="Output path."`

PR #37, opened 2019-03-08
skipping input file if time range is invalid

AliceBalfanz:
when time range is invalid, meaning that the starting time of the data set is later than the stop time, the data set is treated as incorrect and is not processed into the cube.


xcube/api/gen/gen.py
```diff
@@ -150,6 +151,15 @@ def monitor(*args):
     return output_path, status
 
 
+def _check_time_valid(time_range, input_file):
+    if time_range[0] > time_range[1]:
+        print(f"illegal time range: file {input_file}:"
```
forman: Don't print text directly, either use Python logging or warning modules.
Why printing at all? This should be the exception's message string.

xcube/api/gen/gen.py
```diff
@@ -150,6 +151,15 @@ def monitor(*args):
...
+              f" {pd.to_datetime(time_range[0])} > {pd.to_datetime(time_range[1])} ")
+        raise NotImplementedError()
```
forman:
```suggestion
        return None
```

PR #40, opened 2019-03-18
draft developer guide

forman:
Draft developer guide, please review!

### DIFF
--- a/docs/DEV-GUIDE.md
+++ b/docs/DEV-GUIDE.md
@@ -1,5 +1,9 @@
 # Xcube Developer Guide
 
+Version 0.1, draft
+
+*IMPORTANT NOTE: Any changes to this doc must be reviewed by dev-team through pull requests.* 
+
 ## Index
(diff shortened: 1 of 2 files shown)


docs/DEV-GUIDE.md
```diff
@@ -0,0 +1,121 @@
...
+
+*IMPORTANT NOTE: Any changes to this doc must be reviewed by dev-team through pull requests.* 
+
+## Index
+
+TODO
+
+## Adding new functionality
+
+Checklist:
```
TonioF: "Checklist" sounds as if all these steps need to be taken in the specified order. I don't think this is entirely the case here.

forman: Yes! Please help!

forman:
> Good to have this written down!
> 
> Maybe we should also place here the openapi conventions, if there is not another more suitable place.

Yes! Please help!

PR #42, opened 2019-04-18
 - add get_time_range with time and  time_coverage  variables

nyoudjou:
- rename previous definition to get_time_range_v1


environment.yml
```diff
@@ -3,27 +3,27 @@ channels:
...
+  - gdal
+  - matplotlib
+  - netcdf4
+  - numpy
+  - pandas
+  - pyyaml
+  - s3fs
+  - scipy
+  - xarray
+  - zarr
```
forman: This is not acceptable. We need definite environment with fixed package versions. What is the problem with the previous settings?

xcube/api/gen/rbins/iproc.py
```diff
@@ -52,14 +52,50 @@ def get_reprojection_info(self, dataset: xr.Dataset) -> ReprojectionInfo:
                                 xy_crs=CRS_WKT_EPSG_4326,
                                 xy_gcp_step=1)
 
-    def get_time_range(self, dataset: xr.Dataset) -> Tuple[float, float]:
+    def get_time_range_v1(self, dataset: xr.Dataset) -> Tuple[float, float]:
```
forman: What is the use of `get_time_range_v1`? If it is no longer used, please remove it.

xcube/api/gen/snap/iproc.py
```diff
@@ -52,8 +52,13 @@ def get_reprojection_info(self, dataset: xr.Dataset) -> ReprojectionInfo:
...
     def get_time_range(self, dataset: xr.Dataset) -> Tuple[float, float]:
-        t1 = dataset.attrs.get('start_date')
-        t2 = dataset.attrs.get('stop_date') or t1
+
+        if "time_coverage_start" in dataset.attrs:
+            t1 = str(dataset.attrs["time_coverage_start"])
+            t2 = str(dataset.attrs.get("time_coverage_end", t1))
+        else:
+            t1 = dataset.attrs.get('start_date')
+            t2 = dataset.attrs.get('stop_date') or t1
```
forman:
```suggestion
            t2 = dataset.attrs.get('stop_date', t1)
```

xcube/api/gen/rbins/iproc.py
```diff
@@ -52,14 +52,50 @@ def get_reprojection_info(self, dataset: xr.Dataset) -> ReprojectionInfo:
...
         date = dataset.attrs.get('DATE')
         if date is None:
             raise ValueError('illegal L2 input: missing DATE attribute')
         time = dataset.attrs.get('TIME', '1200')
         days_since_1970 = get_time_in_days_since_1970(date + time)
         return days_since_1970, days_since_1970
 
+    def get_time_range(self, dataset: xr.Dataset) -> Tuple[float, float]:
```
forman: This looke very much like the implementation in `DefaultInputProcessor.get_time_range()`.
Try deriving your class from `DefaultInputProcessor` so we don't duplicate code and only need a single test:

```
class RbinsSeviriHighrocSceneInputProcessor(DefaultInputProcessor):
    ...
```

forman:
Also, please note that your branch still created failing builds, see logs, unit tests are still failing. Please fix that too.

xcube/api/gen/rbins/iproc.py
nyoudjou: yes, is not longer used, I will remove it

xcube/api/gen/snap/iproc.py
nyoudjou: I am learning python3 :)

environment.yml
nyoudjou: I had some issue with the install cube on mac os Mojave, I will try to reproduce that

Issue #52, opened 2019-01-21
Improve performance of Time-Series


TonioF:
The performance of the time-series commands is slow, leading to time-outs. This has been observed for all four endpoints.

GET /ts/{dataset}/{variable}/point
POST /ts/{dataset}/{variable}/geometry
POST /ts/{dataset}/{variable}/geometries
POST /ts/{dataset}/{variable}/features

forman:
See also https://github.com/dcs4cop/xcube-viewer/issues/38

forman:
![](https://github.trello.services/images/mini-trello-icon.png) [Address urgent xcube viewer issues](https://trello.com/c/mKJGImqi/31-address-urgent-xcube-viewer-issues)

forman:
At least the TS extraction using `xcube serve` should benefit from #58

forman:
Closing this. Reopen if performance issue persists.

PR #70, opened 2019-05-22
Forman 52 perf tools

forman:
This quite large PR provides some new toys addressing #52 and performance optimisation in general. The actual fix of #52 will follow in another PR. 

* Added new `xcube apply` command that can be used to generate an output cube computed from a Python function 
  that is applied to one or more input cubes. 
  The command is still in development and therefore hidden.
* Added new `xcube timeit` command that can be used to measure the time required for 
  parameterized command invocations, e.g. the `xcube apply` command. 
  The command is still in development and therefore hidden.
* Extended `xcube serve` command by new option `--show` that starts the Xcube viewer on desktop environments in a browser. 
* Refactored out unit test helpers `test.cli.helpers.CliTest` and `test.cli.helpers.CliDataTest`. The latter is used whenever a test requires a NetCDF or ZARR input.


forman:
@dzelge as this PR is now 6 days old, I'll merge it without your review as we have to move on...

PR #71, opened 2019-05-22
xcube documentation

AliceBalfanz:
Updating README, fixing typos


AliceBalfanz:
>     * --show help: Not clear what gets into XCUBE_VIEWER_PATH
> 
>     * --show should allow a paramater that configures the viewer path as an alternative to the env var. Makes dockerisation much easier
> 
>     * xcube-viewer docs needs info about how to install nodejs/yarn
> 
>     * Should we have a default viewer served by tornado as a static website? We have don that in eocdb. Works well

This comment belongs to PR  #70 and needs to be included in the review of it. @dzelge

PR #74, opened 2019-05-28
New Time-Series API

forman:
This PR is still an intermediate step toward resolution of #52:

* refactored out time-series API functions from `webapi.controllers.time_series` into `api.ts` 
* modified time series creation so dask should have an easier job
* distributed contents of `webapi.utils` over `util.geojson`, `util.geom`, `util.timecoords`, then removed module `webapi.utils`


forman:
@TonioF, merging now, as we need to move forward

xcube/webapi/context.py
```diff
@@ -150,15 +150,14 @@ def get_ml_dataset(self, ds_id: str) -> MultiLevelDataset:
         ml_dataset, _ = self._get_dataset_entry(ds_id)
         return ml_dataset
 
-    def get_dataset(self, ds_id: str) -> xr.Dataset:
+    def get_dataset(self, ds_id: str, var_names: Collection[str] = None) -> xr.Dataset:
```
TonioF: From a method with this header, I would expect to get a data set that consists only of the the requested variables (i.e., a variable subset), not that it merely checks whether the variables are present. It is an improvement to the previous method, but at the very least it should be documented.

test/util/test_geojson.py
```diff
@@ -0,0 +1,46 @@
+import unittest
```
TonioF: License is missing.

xcube/util/geom.py
```diff
@@ -0,0 +1,195 @@
...
+def get_box_split_bounds_geometry(lon_min: float, lat_min: float,
+                                  lon_max: float, lat_max: float) -> shapely.geometry.base.BaseGeometry:
+    box_1, box_2 = get_box_split_bounds(lon_min, lat_min, lon_max, lat_max)
+    if box_2 is not None:
+        return shapely.geometry.MultiPolygon(polygons=[shapely.geometry.box(*box_1), shapely.geometry.box(*box_2)])
+    else:
+        return shapely.geometry.box(*box_1)
+
+
+_IVALID_GEOMETRY_MSG = ('geometry must be either a (shapely) geometry object, '
```
TonioF: IVALID -> INVALID

test/util/test_geojson.py
forman: License? Which license, where?

xcube/webapi/context.py
forman: Good point. I'll rename it to `expected_var_names` .

xcube/util/geom.py
forman: Good.

test/util/test_geojson.py
TonioF: The MIT license, just as geom.py, ts.py or ancvar.py have it. I just noticed that I left this comment in the wrong place. It was meant to be placed at the beginning of geojson.py.

PR #109, opened 2019-06-22
Forman 107 server to verify cubes

### LINKED ISSUE #107: xcube server to verify valid cubes

xcube server must verify that configured cubes are valid, so that problems such as #103 are easier to understand address. 

Call `xcube.api.assert_cube(dataset)` after opening datasets.

forman:
Closes #107.

@AliceBalfanz, to verify this PR, please test `xcube serve` with the original (illegal) cube that cause the actual problem. The exception message should now be much cleaner and point you to the actual issue.


forman:
@AliceBalfanz, this is the expected behavior. This PR is about the error message that you get, when an invalid cube is opened. Cubes are never opened, if they are not requested. That's why you don't see any initial error message.

PR #110, opened 2019-06-23
Forman 18 xcube verify

### LINKED ISSUE #19: False cube creation when output name is not set using xcube-genl2c
When the user does not set an output name for the generated data cube when using xcube-genl2c, the output is one output file per input files. This is because the default output name depends on the input file, which changes if there is more than one input file used for the cube generation. The default name should be set to a not changing name.

forman:
Closes #19.

@AliceBalfanz, @dzelge  in the future, please apply the new tool `xcube verify INPUT` on each generated data cube.

@gunbra32, FYI.


AliceBalfanz:
closes #107

Issue #64, opened 2019-05-09
xcube gen to force chronological order


forman:
**Is your feature request related to a problem? Please describe.**

`xcube gen` can only append time slices. But we also need to prepend and/or insert time slices into existing cubes so we maintain chronological order.

**Describe the solution you'd like**

Add new `xcube gen` option `--forcechron`.

**Describe alternatives you've considered**

See #47.

**Additional context**

See

*  [Continuously extending Zarr datasets](https://medium.com/pangeo/continuously-extending-zarr-datasets-c54fbad3967d)
* #45

**EDIT**

See also

* ZARR [parallel append](https://github.com/zarr-developers/zarr/issues/75)

forman:
Merged already. Note, overwrite at time step is missing....

Issue #45, opened 2019-04-30
xcube gen generates duplicates in time and unsorted time dimension for S2+


gunbra32:
**Describe the bug**
xcube gen generates two identical time slices with S2+ plugin. In the input data there is only one input file. In addition, the time dimension is not sorted. 

**To Reproduce**
The cube has been produced in xcube-gen with the the following command. 
nohup xcube gen --append -v CHL_GILERSON2010_GLOBAL,KdPAR,SPM_VITOnir_SCHELDT,TUR_NECHAD2009_GLOBAL_665 -c /home/xcube/projects/xcube-services/dcs4cop/xcube-gen-configs/dcs4cop-gen_BC_config_S2.yml /data/EOdata/related/DCS4COP/cube_input/VITO/2017/*/*.nc > nohup.out &
The resulting zarr file is still on xcube-gen at /home/xcube/xcube-output/ dcs4cop-bc-s2-sns-l2c-v1.zarr
Note that this issue might be related to the fact that the generation has been interrupted and started again with the same output name. 

**Screenshots**
![image](https://user-images.githubusercontent.com/7138923/56975208-97df1700-6b70-11e9-85af-a6fcc7e7caa2.png)

forman:
Maybe a duplicate of #33 ?

forman:
![](https://github.trello.services/images/mini-trello-icon.png) [Generate S2-Plus data cubes for generic use case](https://trello.com/c/OGLyedMK/2-generate-s2-plus-data-cubes-for-generic-use-case)

forman:
![](https://github.trello.services/images/mini-trello-icon.png) [Generate S2-Plus data cubes for generic use case](https://trello.com/c/OGLyedMK/2-generate-s2-plus-data-cubes-for-generic-use-case)

forman:
Merged.

PR #124, opened 2019-07-16
force chronology

AliceBalfanz:
This PR inplements changes in the data cube creation in zarr format. The time stamp of each input file is checked and depending on time chronology either appended or inserted at the correct place. When a time stamp is already included in the data cube, the input file is skipped. 
This fixes the problem of breaking time chronology through appending time stamps which should be inserted, issue #64 and issue #45.


AliceBalfanz:
Found a bug, closing pull request untill fixed.

AliceBalfanz:
The comparison, whether a time stamp is already included  in an existing data cube is at the moment very late, therefore the data set is processed already and then the decision comes about dropping it due to being a duplicate. @forman it would be great to get some feedback, how to solve this issue earlier and making the process of appending and merging faster. A difficulty is, that the time stamp gets rounded while appending it to zarr, this means the center of the start and end time of the input netcdf cannot be compared to the existing  time stamps in the data cube.

xcube/util/zarrinsert.py
```diff
@@ -0,0 +1,126 @@
...
+    mask = np.equal(ds.time.values, ds_single.time.values)
+    ds.close()
+    ds_single.close()
+    if not mask.any():
+        return True
+    else:
+        return False
+
+
+def _rename_file(path_to_ds, old_index, new_time_i):
```
forman: This implementation does not recognize if nested zarr is used. See https://zarr.readthedocs.io/en/stable/api/storage.html#zarr.storage.NestedDirectoryStore

xcube/util/zarrinsert.py
```diff
@@ -0,0 +1,126 @@
...
+                        os.rename(os.path.join(path, filename), os.path.join(path, str(new_time_i)))
+
+
+def _copy_into_target(src_path, dst_path, src_index):
+    """Copy the files with the new time stamp into the existing zarr directory."""
+    ds = xr.open_zarr(src_path)
+    variables = ds.variables
+    ds.close()
+    for v in variables:
+        if (v != 'lat') and (v != 'lon') and (v != 'lat_bnds') and (v != 'lon_bnds'):
```
forman: This is not correct, because any non-time coordinate variables must be excluded. So exclude all coordinate variables that are not time-related:

```
    time_var = ds['time']
    time_bnds_var_name = time_var.attrs.get('bounds', 'time_bnds')
    excluded_var_names = set(ds.coord_vars).difference({'time', time_bnds_var_name})
    ...
       if v not in excluded_var_names:
           ...
```

If it was correct: 
- Remove redundant parentheses
- Better use `in`: `if v not in {'lon', 'lat', 'lon_bnds', 'lat_bnds'}:`

xcube/util/zarrinsert.py
```diff
@@ -0,0 +1,126 @@
...
+    ds_single.close()
+    if not mask.any():
+        return True
+    else:
+        return False
+
+
+def _rename_file(path_to_ds, old_index, new_time_i):
+    """Renaming files within the directories according to new time index."""
+    ds = xr.open_zarr(path_to_ds)
```
forman: Use `with`.

xcube/util/zarrinsert.py
```diff
@@ -0,0 +1,126 @@
...
+    variables = ds.variables
+    ds.close()
```
forman: Don't use dataset data after closing datasets or make sure you deep copy the data.

xcube/util/zarrinsert.py
```diff
@@ -0,0 +1,126 @@
...
+    for v in variables:
+        if (v != 'lat') and (v != 'lon') and (v != 'lat_bnds') and (v != 'lon_bnds'):
+            path = os.path.join(path_to_ds, v)
+            for root, dirs, files in os.walk(path):
+                for filename in files:
+                    parts1 = filename.split('.', 1)[0]
+                    if parts1 == (str(old_index)) and (v != "time"):
```
forman: What about `time_bnds`?

xcube/util/zarrinsert.py
```diff
@@ -0,0 +1,126 @@
...
+        for i in reversed(range(new_idx, ds.time.shape[0])):
+            _rename_file(dst_path, i, (i + 1))
+        # Final step: copy the single time stamp files into the destination zarr and adjusting .zarray to the change.
+        _copy_into_target(src_path, dst_path, new_idx)
+        return True
+    else:
+        return False
+
+
+def _check_if_unique(src_path, dst_path):
```
forman: Make function names more explicit, e.g. `_can_merge_dataset`

xcube/api/gen/gen.py
```diff
@@ -254,10 +256,36 @@ def step8(dataset):
         if append_mode and os.path.exists(output_path):
             # noinspection PyShadowingNames
             def step9(dataset):
-                output_writer.append(dataset, output_path, **output_writer_params)
+                global _APPEND_DS_TO_DC
+                if output_path.endswith('.nc'):
+                    output_writer.append(dataset, output_path, **output_writer_params)
+                elif output_path.endswith('.zarr'):
+                    _APPEND_DS_TO_DC = check_append_or_insert(time_range, output_path)
+                    if _APPEND_DS_TO_DC is True:
```
forman: Don't use `is` if right side expression is `bool`.

xcube/api/gen/gen.py
```diff
@@ -254,10 +256,36 @@ def step8(dataset):
...
+                        output_writer.append(dataset, output_path, **output_writer_params)
+                    elif _APPEND_DS_TO_DC is False:
```
forman: Don't use `is` if right side expression is `bool`.

xcube/api/gen/gen.py
```diff
@@ -254,10 +256,36 @@ def step8(dataset):
...
+                        tempdir_for_merging = os.path.join(os.path.dirname(__file__), '..', '..', '..',
```
forman: Oh nooo, this is really bad!

Use OS-specific temporary location, use https://docs.python.org/3.7/library/tempfile.html

xcube/api/gen/gen.py
```diff
@@ -254,10 +256,36 @@ def step8(dataset):
...
+                                                           'tempdir_for_merging')
+                        try:
+                            os.mkdir(tempdir_for_merging)
+                        except OSError:
```
forman: This will always be raised if `xcube` is run from a docker.

xcube/api/gen/gen.py
```diff
@@ -254,10 +256,36 @@ def step8(dataset):
...
+                        merged_data_set = merge_single_zarr_into_destination_zarr(temp_output_path, output_path)
+                        rimraf(tempdir_for_merging)
+                        if not merged_data_set:
+                            monitor('Time Stamp of input data set is already existing in data cube: skipping...')
+                            return False
+
                 return dataset
 
-            steps.append((step9, f'appending to {output_path}'))
+            steps.append((step9, f'appending or merging to {output_path}'))
```
forman:
```suggestion
            steps.append((step9, f'appending to or inserting into {output_path}'))
```

or use `_APPEND_DS_TO_DC` (as local variable of course) to decide which text to print.

xcube/cli/gen.py
```diff
@@ -68,8 +68,10 @@
 @click.option('--append', '-a', is_flag=True,
               help='Append successive outputs.')
```
forman: Now, we should really think about dropping `--append`. Append-mode should be default, that is, if a cube exists append new slices, if not create one.

It would now be better to introduce a new option `--new` to force creating new, non-existent data cubes. If `--new` is given and the cube exists, the command then fails.

xcube/util/zarrinsert.py
```diff
@@ -0,0 +1,126 @@
...
+        t_center = t1
+    ds = xr.open_zarr(output_path)
+    if np.greater(t_center, get_time_in_days_since_1970(ds.time.values[-1])):
+        # append modus
+        return True
+    else:
+        return False
+
+
+def merge_single_zarr_into_destination_zarr(src_path, dst_path):
```
forman: Name `single_zarr` is misleading.

xcube/api/gen/gen.py
```diff
@@ -254,10 +256,36 @@ def step8(dataset):
...
+                                                           'tempdir_for_merging')
+                        try:
+                            os.mkdir(tempdir_for_merging)
+                        except OSError:
+                            monitor("Creation of the directory %s failed" % tempdir_for_merging)
+                        else:
+                            monitor("Successfully created the directory %s for inserting "
+                                    "time stamp " % tempdir_for_merging)
+                        temp_output_path = os.path.join(tempdir_for_merging,
+                                                        input_file.replace("/", "_")[:-3] + '.zarr')
```
forman: What does this magic do?

xcube/api/gen/gen.py
```diff
@@ -254,10 +256,36 @@ def step8(dataset):
...
+                        output_writer.write(dataset, temp_output_path, **output_writer_params)
+                        merged_data_set = merge_single_zarr_into_destination_zarr(temp_output_path, output_path)
+                        rimraf(tempdir_for_merging)
+                        if not merged_data_set:
+                            monitor('Time Stamp of input data set is already existing in data cube: skipping...')
```
forman: Please fix text: 
```suggestion
                            monitor('Time stamp of input already exists in output: skipping it...')
```

Note, we should let the monitor do real logging instead of printing to stdout.

xcube/util/zarrinsert.py
```diff
@@ -0,0 +1,126 @@
...
+from xcube.util.timecoord import get_time_in_days_since_1970
+
+
+def check_append_or_insert(time_range, output_path):
+    t1, t2 = time_range
+    if t1 != t2:
+        t_center = (t1 + t2) / 2
+    else:
+        t_center = t1
+    ds = xr.open_zarr(output_path)
```
forman: Use `with` so `ds` is closed.

xcube/util/zarrinsert.py
```diff
@@ -0,0 +1,126 @@
...
+import json
+import numpy as np
+import os
+import shutil
+import xarray as xr
+
+from xcube.util.timecoord import get_time_in_days_since_1970
+
+
+def check_append_or_insert(time_range, output_path):
```
forman: Add types.

xcube/util/zarrinsert.py
```diff
@@ -0,0 +1,126 @@
...
+            for root, dirs, files in os.walk(path):
+                for filename in files:
+                    parts1 = filename.split('.', 1)[0]
+                    if parts1 == str(src_index):
+                        shutil.copyfile((os.path.join(src_path, v, filename)), (os.path.join(dst_path, v, filename)))
+            _adjust_zarray(dst_path, v)
+
+
+def _adjust_zarray(dst_path, variable):
+    """Changing the shape for time in the .zarray file."""
```
forman: The function does not look very "time"-specific.

The name `_adjust_zarray` does not tell what it actually does. Use better name e.g. `_expand_1d_var_shape`

xcube/util/zarrinsert.py
AliceBalfanz: instead use `input_zarr`. and for destination zar use `output_zarr`.

AliceBalfanz:
Closing this pull request due to different implementation approach by @forman which was merged into master by PR #137, not deleting branch in order to keep the solution as fall back solution.

Issue #3, opened 2018-11-12
Store cubegen parameterisation


forman:
Store all xcube-genl2c, -genl3 parameters and source file names in metadata. Use CF fields `history` and `sources`.

PR #127, opened 2019-07-25
Adding cubegen parametrisation as "history" to cube attributes 

AliceBalfanz:
This pull request adds "history" to the cube's attribute metadata containing the parametrisation for cube generation. This relates to issue #3 .


AliceBalfanz:
According to the [CF Convention](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.pdf#%5B%7B%22num%22%3A197%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C0%2C488.93%2Cnull%5D), 
`histroy` provides a list of performed changes on the original data and `source ` gives information about the original input data. 

> What about CF sources global attribute?

I understand `source` this as e.g. 'Sentinel-3 OLCI L2 surface observation' and not as a list of input files. This is why I would like to store the parameters for the cube generation not in `source ` but in `history`.  The CF convention document recomends to add a line with a time stamp for each time, the data set is changed. Due to our case of growing data cube, this would result in a very long list of input files, so maybe better not include the input file list into the history? 
However it is interesting to store e.g. the resampling method in the global attributes as well as the pixel expressions in order to see, how valid pixels are defined. 
The CF convention does offer an attribute called `comment` which may be used as well to add information about the data production. Maybe this would be the more suitable place to store the cube gen parameters. 

> Why do you change the user's config that is supposed to only contain user-provided information?

I do not understand this remark. The code does not change the users config, it just adds the configuration parameters to the metadata global atributes. And in order to do so, it takes the users config parameters and adds them. 

> Please describe your solution first. How does it resolve #3?

The solution at the moment is, that the output_metadata is updated by adding the parameters used for the cube generation and are written to the data cubes global attribute `history`. It does not delete information provided by the user using the `output_metadata` parameter within the configuration file, it only adds the data. This is how I understand, issue #3 should be solved.

forman:
https://help.ceda.ac.uk/article/4432-ceda-cf-examples

forman:
> I do not understand this remark. The code does not change the users config, it just adds the configuration parameters to the metadata global atributes. ...

It is a software design issue. It is just not the right place to put it into the global attributes in the config. Instead add the new global attributes before the output is written.

AliceBalfanz:
Closing for the moment, because still working on it, but paused for the moment.

PR #131, opened 2019-07-26
Forman 130 lazy load dataset details

### LINKED ISSUE #130: Allow lazy-loading of dataset details

We currently include all GeoJSON features associated with a dataset in the dataset's metadata when using the WebAPI `datasets?details=1`. This may cause server's to respond with very high latency as a lot of coordinate data may be associated with datasets e.g. all Swedish lakes assoc with a Sweden datacube. See https://github.com/dcs4cop/xcube-viewer/issues/61

Allow clients to lazy-load GeoJSON features and also a dimension's coordinate arrays, which may be very large, hence extend the WebAPI by new functions.

forman:
After merging this PR, `xcube serve`  will behave differently: 

* The response of `datasets?details=1` will now contain dataset descriptors, whose `placeGroups` GeoJSON feature collections will **no longer** have a `features` field. 
* Clients must therefore call `datasets/{dataset}/places/{placeGroupId}` to explicitly receive a place group including all GeoJSON features.

**WARNING**: Deploying new server instances will require to also deploy updated xcube-viewer instances. See https://github.com/dcs4cop/xcube-viewer/pull/65 and https://github.com/dcs4cop/xcube-viewer/issues/61

Closes #130.


forman:
> Should these changes be included into the CHANGES.md? 

Yes. It is an improvement.

> Because there is an addition in the possibilities of regesting places?

I don't understand this.

AliceBalfanz:
> > Because there is an addition in the possibilities of regesting places?
> 
> I don't understand this.

Sorry, very bad English... What I meant to write: Shouldn't the changes be included in the CHANGES.md because amongst others there is a change of how places are requested, and the user should be aware of this change. But you already answered, that it should be included in the changes.md :)

PR #153, opened 2019-08-14
Fix TODO: Adjust global attrs

forman:
This PR fixes the "# TODO (forman): Adjust global attrs" in function `clip_dataset_by_geometry()`. For the fix:

* moved xcube.api.update to xcube.util.update
* exported xcube.util.update.* in xcube.api
* split attribute update function into individual spatial / temporal updates


xcube/util/update.py
```diff
@@ -26,76 +26,82 @@
...
+    Update spatio-temporal CF/THREDDS attributes given *dataset*.
 
-    :param dataset: A dataset.
-    :param var_name_to_props: List of tuples of the form (variable name, properties dictionary).
-    :return: A new dataset.
+    :param dataset: The dataset.
+    :param global_attrs: Optional global attributes.
+    :param update_existing: If ``True``, any existing attributes will be updated.
+    :param in_place: If ``True``, *dataset* will be modified in place and returned.
+    :return: A new dataset, if *in_place* if ``False`` (default), else the passed and modified *dataset*.
```
AliceBalfanz:
```suggestion
    :return: A new dataset, if *in_place* is ``False`` (default), else the passed and modified *dataset*.
```

xcube/util/update.py
```diff
@@ -26,76 +26,82 @@
...
     """
-    if not var_name_to_props:
-        return dataset
+    if not in_place:
```
AliceBalfanz: I do not understand when `in_place` would be true. Could you explain?

xcube/util/update.py
```diff
@@ -26,76 +26,82 @@
...
+                                 update_existing: bool = False,
+                                 in_place: bool = False) -> xr.Dataset:
+    """
+    Update spatial CF/THREDDS attributes of given *dataset*.
 
-    return dataset
+    :param dataset: The dataset.
+    :param update_existing: If ``True``, any existing attributes will be updated.
+    :param in_place: If ``True``, *dataset* will be modified in place and returned.
+    :return: A new dataset, if *in_place* if ``False`` (default), else the passed and modified *dataset*.
```
AliceBalfanz:
```suggestion
    :return: A new dataset, if *in_place* is ``False`` (default), else the passed and modified *dataset*.
```

xcube/util/update.py
```diff
@@ -26,76 +26,82 @@
...
     """
-    Update the global CF/THREDDS attributes of given *dataset*.
+    Update temporal CF/THREDDS attributes of given *dataset*.
 
     :param dataset: The dataset.
-    :param output_metadata: Extra metadata.
-    :return: A new dataset.
+    :param update_existing: If ``True``, any existing attributes will be updated.
+    :param in_place: If ``True``, *dataset* will be modified in place and returned.
+    :return: A new dataset, if *in_place* if ``False`` (default), else the passed and modified *dataset*.
```
AliceBalfanz:
```suggestion
    :return: A new dataset, if *in_place* is ``False`` (default), else the passed and modified *dataset*.
```

xcube/util/update.py
AliceBalfanz: Sorry, I was to quick with my question. I see now where it is set to true.

PR #150, opened 2019-08-13
Forman 148 subset by geometry

forman:
* Added new spatial dataset operations to `xcube.api`:
  * `mask_dataset_by_geometry(dataset, geometry_like)` clip and mask a dataset by geometry
  * `clip_dataset_by_geometry(dataset, geometry_like)` just clip a dataset by geometry
  * `convert_geometry(geometry_like)` convert a geometry-like into a shapely geometry
* Renamed type `Geometry` into `GeometryLike`
* Fixed clipping of datasets (clip bounds were by 1 too large in x,y directions)
* Fixed conversion of geometry-like objects if GeoJSON dicts were passed
* Added a lot of missing tests, adapted tests

Closes #148

@AliceBalfanz to merge this PR asap

Sorry, the PR became quite large.

PR #156, opened 2019-08-19
Forman mask dataset by geom exclude vars

forman:
`mask_dataset_by_geometry()` has new keyword argument `excluded_vars` so we can avoid flag variables becoming masked (which doesn't make sense).

This adds a missing capability to #150


forman:
@dzelge ?

dzelge:
Approved

PR #169, opened 2019-09-13
Dividing former example notebook in smaller ones

AliceBalfanz:
Dividing the large notebook into several smaller ones, notebooks for timeseries extraction  and point extraction are missing at the moment, but will follow next week.


AliceBalfanz:
Closing this, because a new example cube is needed, based on which the examples will be changed.

PR #172, opened 2019-09-17
Dzelge 94 warnings 3rd party

dzelge:
When accepting this PR the number of warnings during testing will be significantly reduced. Two
warnings remain that will disappear when we change to Python 3.8.

- Some warnings are supressed as they cannot be entirely avoided (e.g. numpy warnings for all NaN slices)
- Python warnings are now caught and checked by self.assertWarns (mainly in plugin.py)
- Some deprecation warnings due to default changes in xarray have been fixed

The foloowing warnings are safely suppressed but could/should get attention:
- matplotlib: tight_layout warning
- time unit warning when writing to netcdf

### DIFF
--- a/test/api/gen/default/test_gen.py
+++ b/test/api/gen/default/test_gen.py
@@ -44,7 +44,8 @@ class DefaultProcessTest(unittest.TestCase):
             self.assertTrue('\nstep 8 of 8: creating input slice in l2c.nc...\n' in output)
             self.assertTrue('\nstep 8 of 8: appending input slice to l2c.nc...\n' in output)
 
-            # Ignoring warning Variable 'time' has datetime type and a bounds variable but time.encoding does not have units specified...
+            # Ignoring warning Variable 'time' has datetime type and a bounds variable but time.encoding
+            # does not have units specified...
             # TODO: Find out why this happens
             for w in warning.warnings:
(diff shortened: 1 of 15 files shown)


dzelge:
> 
> Why are we now asserting in many tests that a certain warning is raised?
> 

Lets discuss, but the documentation advises explictely not ignore warnings when running tests. I would ,therefore, prefer ignoring them also during runtime. Some of the warnings can be quite useful like division by zero warnings.

forman:
Ok, but that's not my point. I'm fine with having warnings enabled during test, however we do not *expect* them, therefore we don't need to *assert* them.

Users are confused by warnings during execution of xcube CLI commands and when using the API in their own programs or notebooks.

So we may have a global xcube CLI flag `-w,--warn`  to enable warnings, by default they are off.
For the API we can introduce a function for that purpose, but first see, how this issue is handled in xarray, numpy etc.

dzelge:
> Users are confused by warnings during execution of xcube CLI commands and when using the API in their own programs or notebooks.
> 
> So we may have a global xcube CLI flag `-w,--warn` to enable warnings, by default they are off.
> For the API we can introduce a function for that purpose, but first see, how this issue is handled in xarray, numpy etc.

Agree. Had actually the same idea.

forman:
Ok, so we turn this into a feature requests! Can you please post an issue?

PR #180, opened 2019-09-20
Dzelge loosening environment

dzelge:
When accepting this PR xcube:

- Will have a loser build environment which includes python 3.6
- Will have a Time series fix that allows to use xarray 0.13
- Omits the IndexStore test. That should and will be a different PR


forman:
Good! However, please add entry to CHANGES.md saying that versions of package requirements have been relaxed. After doing so you can merge right away.

PR #182, opened 2019-09-24
Alicja 106 notebooks changes

AliceBalfanz:
I changed the notebooks, made shorter ones. 
Changed the cube which is used for the demo, it is now SST for the North Sea.


examples/notebooks/show_remote_cubes.py
```diff
@@ -0,0 +1,34 @@
...
+            chunking = ds[var_list[0]].shape
+            sd= pd.to_datetime(str(ds.time.values[0])) 
+            start_date = sd.strftime('%Y-%m-%d')
+            ed= pd.to_datetime(str(ds.time.values[-1])) 
+            end_date = ed.strftime('%Y-%m-%d')
+            chunksize = ds[var_list[0]].data.chunksize
+            df = df.append({'cube_name': filename, 'chunksize (time, lat, lon)': chunksize, 
+                            'var_nums':len(var_list), 'variables':(str(var_list)).replace('[','').replace(']','').replace("'",""), 
+                            'start_date':start_date, 'end_date':end_date}, ignore_index=True)
+    return df
```
dzelge: Add a new line

examples/notebooks/show_remote_cubes.py
```diff
@@ -0,0 +1,34 @@
+import os
+import pandas as pd
+import s3fs
+from xcube.api import read_cube
+
+def show_remote_cubes():
```
dzelge: - Add options allowing users to use this function for other buckets and endpoints.
- Could/should be integral part of xcube?

examples/notebooks/show_remote_demo_cubes.ipynb
```diff
@@ -0,0 +1,277 @@
+{
```
dzelge: It is a cool notebook (I should say that as we use this in ESDL). However, this is more an example of using S3 
not of xcube. I note later that this could be integral part of xcube. Then it would become an xcube example. 

You could open a feature reqeust for this.

examples/notebooks/make_subset_cube.ipynb
```diff
@@ -0,0 +1,398 @@
+{
```
dzelge: Nice. Nice colours. Maybe you need not print the cube as you have done that before. But you might have good reasons to do that,

examples/notebooks/show_remote_demo_cubes.ipynb
AliceBalfanz: @forman What do you think about Helges suggestion?

PR #185, opened 2019-09-24
Forman 181 no metadata written

### LINKED ISSUE #181: Metadata passed via config not written to cube
**Describe the bug**
The metadata specified in a config file which is passed to generate an xcube dataset is not used by xcube gen to write it into the metadata of the cube. 

**To Reproduce**
1. Go to the xcube docu [example section](https://xcube.readthedocs.io/en/latest/examples/xcube_gen.html#sea-surface-temperature-anomalies-over-the-global-ocean)
2. Generate the example data cube as described
3. Take a look at the metadata of the generated cube using `xcube dump `
grafik

4. Compare to the output_metadata specified in the [config file](https://github.com/dcs4cop/xcube/blob/master/examples/gen/config_files/dcs4cop-gen_BC_config_CMEMS.yml) . 

**Expected behavior**
The specified output_metadata should be included in the metadata of the cube and should appear when using xcube dump to view the metadata. 
This has been working before.

forman:
Closes #181

Note: The core of this fix is the find that `dataset.to_zarr(store, model='a', append_dim='time')` will replace global attributes in `store` by global attributes in `dataset`. I guess this is an xarray or zarr bug or at least an unexepected behaviour in my point of view. In append mode, attributes should updated rather than being replaced.


forman:
Review performed.

Issue #167, opened 2019-09-12
Cube computations


forman:
**Is your feature request related to a problem? Please describe.**

In order to compute target variables from one or more source variables it is often sufficient and efficient to perform array operations using standard Python expression syntax. Such expression may involve cube variables (`xarray.DataArray` instances), other numpy-array-like objects, or scalars (dimension broadcasting may takes place) combined by standard operators and numpy-ufuncs. This will ensure that out-of-core computation can take place and dask can care for the parallelisation of chunk computations.

However, this approach will only scale to a certain point. If the algorithm to be implemented is non-trivial (large number of complex array operations), the overhead caused by dask chunk management and xarray metadata management is no longer neglectable. For such cases, xarray provides the [`xarray.apply_ufunc`](http://xarray.pydata.org/en/stable/generated/xarray.apply_ufunc.html) function that allows applying a [numpy-ufunc-conformant](https://docs.scipy.org/doc/numpy/reference/ufuncs.html) function to one or more source chunks to compute a target chunk. This is a very nice feature, however it has the drawback, that the user-provided ufunc will not receive the actual coordinate labels of the sub-region that is currently being computed. But many many (EO) algorithms are dependent on time, latitude, and/or longitude (algorithms utilising FFT require time coordinates, other intermediate variables may be latitude-dependent).  In addition, it is not trivial to configure `xarray.apply_ufunc` for algorithms that reduce dimensions or generate new ones.

**Describe the solution you'd like**

Provide an xcube wrapper for `xarray.apply_ufunc` that passes coordinates of the sub-region to the user function. In addition, allow specifying the new output dimensions in an intuitive way using the dimension names and sizes.

forman:
Related branch: https://github.com/dcs4cop/xcube/tree/forman-167-cube_computations

forman:
Done.

PR #205, opened 2019-10-16
Forman 167 cube computations

forman:
This is a big one:

### New

* Added new `xcube compute` CLI command and `xcube.api.compute_cube()` API 
  function that can be used to generate an output cube computed from a Python
  function that is applied to one or more input cubes. Replaces the formerly 
  hidden `xcube apply` command. (#167)
* Added new `xcube.api.CubeStore` (to be used by ESDL PP code later!)
* Added new `xcube.api.CubeSchema`

### Incompatible changes

The following changes introduce incompatibilities with former xcube 0.2.x 
versions. 

* `xcube.api.compute_dataset()` function has been renamed to 
  `xcube.api.evaluate_dataset()`. This has been done in order avoid confusion
  with new API function `xcube.api.compute_cube()`.


forman:
> Could you please check, why the Travis Tests are failing?
> Other than that, I do agree :)

Yepp. However, because of a GDAL update, also `master` currently fails.

PR #220, opened 2019-11-06
Forman 212 resample bug

### LINKED ISSUE #212: Resample .zarr Cube
**Describe the bug**
Error during the xcube resample for Sentinel-2 dataset.

**To Reproduce**
1. Generate the Sentinel-2 MSI irregular time cube using the `xcube gen`.

To generate it I took the netCDF files (tiles 31UET, 31UES, 31UDT, 31UDS).

**I use the config parameters:**
TICKET01_01

The **output_size** and **output_region** I took from `xcube grid abox`:
TICKET01_02

**The .zarr cube was created with no errors.**
ticket01_06

**2. Run the `xcube resample` with the command line:**
ticket01_03

**HERE IS THE ERROR MESSAGE:**
xcube_ticket01_resample_issue_with_chunk

**Run again the `xcube resample` but now with output as netCDF4:**
ticket01_04

**With output as netCDF4 the resample successfully finished.**
ticket01_05

**Expected behavior**
I was expected to resample using as output the .zarr format (netCDF4 is too big and too slow).

**Screenshots**
Added during the text.

**Desktop (please complete the following information):**
 - OS: Linux Ubuntu 18.04
 - Browser: Firefox (not necessary)
 - Version: concerning the xcube version I updated the xcube and xcube-gen-bc before start (30/10/2019)

**Additional context**
I used the same `xcube resample` before for the Sentinel-3 (full North Sea) and it worked perfect to generate the 1D resample cube in .zarr format. **I think my problem is due the chunk option.**
What should I do to edit the chunk size to a *ideal match*.

Thank you for the help!!

forman:
* `xcube resample` now correctly re-chunks its output. By default, chunking of the 
  `time` dimension is set to one. (#212)
* refactored out new function `xcube.core.update.update_dataset_chunk_encoding()`

Closes #212

### DIFF
--- a/xcube/core/resample.py
+++ b/xcube/core/resample.py
@@ -43,10 +43,11 @@ def resample_in_time(cube: xr.Dataset,
     Resample a xcube dataset in the time dimension.
 
     :param cube: The xcube dataset.
-    :param frequency: Resampling frequency.
+    :param frequency: Temporal aggregation frequency. Use format "<count><offset>"
+        "where <offset> is one of 'H', 'D', 'W', 'M', 'Q', 'Y'.
     :param method: Resampling method or sequence of resampling methods.
     :param offset: Offset used to adjust the resampled time labels.
-        Some pandas date offset strings are supported.
+        Uses same syntax as *frequency*.
     :param base: For frequencies that evenly subdivide 1 day, the "origin" of the
         aggregated intervals. For example, for '24H' frequency, base could range from 0 through 23.
     :param time_chunk_size: If not None, the chunk size to be used for the "time" dimension.
--- a/xcube/core/resample.py
+++ b/xcube/core/resample.py
@@ -23,32 +23,43 @@ from typing import Dict, Any, Sequence, Union
 
 import xarray as xr
 
+from xcube.core.schema import CubeSchema
 from xcube.core.select import select_vars
+from xcube.core.verify import assert_cube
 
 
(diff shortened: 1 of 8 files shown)


xcube/core/resample.py
```diff
@@ -23,32 +23,43 @@
...
     """
     Resample a xcube dataset in the time dimension.
 
     :param cube: The xcube dataset.
     :param frequency: Resampling frequency.
     :param method: Resampling method or sequence of resampling methods.
-    :param offset: Offset used to adjust the resampled time labels. Some pandas date offset strings are supported.
-    :param base: Resampling method.
+    :param offset: Offset used to adjust the resampled time labels.
+        Some pandas date offset strings are supported.
```
dzelge: Would it be possible to add a link to a doc that lists valid offset strings are supported?

xcube/core/resample.py
```diff
@@ -23,32 +23,43 @@
...
+    :param base: For frequencies that evenly subdivide 1 day, the "origin" of the
+        aggregated intervals. For example, for '24H' frequency, base could range from 0 through 23.
```
dzelge: Link to a doc that lists options?

xcube/core/resample.py
forman: Its better saying that it uses same syntax as `frequency`.  Done.

PR #221, opened 2019-11-06
Input list sorted based on time stamp

AliceBalfanz:
This PR implements sorting of the input list based on the dataset time stamps by default. The order of the input file list may be kept by passing `--no_sort`. The parameter `--sort` is deprecated with this PR.

PR #223, opened 2019-11-08
modified presorting of file paths

AliceBalfanz:
This PR is a continuation of PR #221 . It was to specific and was only working with the default input processor. now new abstract method is added, which gets the time stamp based on the input processor.

### DIFF
--- a/xcube/core/gen/iproc.py
+++ b/xcube/core/gen/iproc.py
@@ -251,21 +251,6 @@ class DefaultInputProcessor(XYInputProcessor):
     def input_reader(self) -> str:
         return self._input_reader
 
-    # def get_time_for_sorting(self, dataset: xr.Dataset) -> Optional[str]:
-    #     time_string = None
-    #     if "time" in dataset:
-    #         time_string = str(dataset.time[0].values)
-    #     if time_string is None:
-    #         if "time_coverage_start" in dataset.attrs:
-    #             time_string = str(dataset.attrs["time_coverage_start"])
-    #         elif "time_start" in dataset.attrs:
-    #             time_string = str(dataset.attrs["time_start"])
-    #         elif "start_time" in dataset.attrs:
-    #             time_string = str(dataset.attrs["start_time"])
-    #     if time_string is None:
-    #         raise ValueError("invalid input: missing time coverage information in dataset")
-    #     return time_string
-
     def pre_process(self, dataset: xr.Dataset) -> xr.Dataset:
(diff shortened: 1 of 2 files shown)


xcube/core/gen/iproc.py
```diff
@@ -101,6 +101,13 @@ def input_reader_params(self) -> dict:
         return dict()
 
     @abstractmethod
+    def get_time_for_sorting(self, dataset: xr.Dataset) -> Optional[str]:
+        """
+        Return a string of the datasets time for presorting the input list.
+        :param dataset: The dataset.
+        :return: The time string of the dataset or None.
+        """
+
```
forman: Why this new method? For sorting, `get_time_range` should be sufficient.
If there is a special need for this method, why it is abstract and why doesn't it have default implementation, e.g.

    def get_time_for_sorting(self, dataset: xr.Dataset) -> Optional[str]:
        t1, t2 = self.get_time_range()
        return (t1 + t2) / 2

xcube/core/gen/iproc.py
```diff
@@ -252,6 +259,21 @@ def configure(self, input_reader: str = 'netcdf4'):
...
+            if "time_coverage_start" in dataset.attrs:
+                time_string = str(dataset.attrs["time_coverage_start"])
+            elif "time_start" in dataset.attrs:
+                time_string = str(dataset.attrs["time_start"])
+            elif "start_time" in dataset.attrs:
+                time_string = str(dataset.attrs["start_time"])
+        if time_string is None:
+            raise ValueError("invalid input: missing time coverage information in dataset")
+        return time_string
+
```
forman: Redundant. See above.

xcube/core/gen/iproc.py
```diff
@@ -252,6 +251,21 @@ def configure(self, input_reader: str = 'netcdf4'):
...
+    #         if "time_coverage_start" in dataset.attrs:
+    #             time_string = str(dataset.attrs["time_coverage_start"])
+    #         elif "time_start" in dataset.attrs:
+    #             time_string = str(dataset.attrs["time_start"])
+    #         elif "start_time" in dataset.attrs:
+    #             time_string = str(dataset.attrs["start_time"])
+    #     if time_string is None:
+    #         raise ValueError("invalid input: missing time coverage information in dataset")
+    #     return time_string
+
```
forman: Remove comment or comment the comment by saying why it should stay.

Issue #112, opened 2019-06-23
Allow for other common spatial CRS


forman:
**Is your feature request related to a problem? Please describe.**

xcube currently does only support spatial grids defined on the spatial CRS [EPSG:4326](https://spatialreference.org/ref/epsg/wgs-84/) (Plate Carrée projection).

Many applications require common CRSs based on Transverse Mercator or Azimuthal Lambert projections usually because their cubes are regional and medium to high spatial resolution.  

**Describe the solution you'd like**

The CRS should be encoded in the cube's global attributes following [CF conventions](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/ch05s06.html). A special attribute `proj4_params` will contain a [PROJ4](http://proj.maptools.org/gen_parms.html)-compatible parameter string.

* For any CRS that uses decimal degrees as coordinate units, the spatial dimensions and coordinate variables will continued to be named `lon` and `lat`.
* For other CRSs, the spatial dimensions and coordinate variables will be named `x` and `y`.

**Describe alternatives you've considered**

On-the-fly reprojection by `xcube serve`.

**Additional context**

Urgently required by DCFS project where we need to support Level-1 and Level-2 EO data for multiple high-resolution sensors.

See also 

* [CF-Conventions on Coordinate Systems](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/ch05s06.html)
* [CF-1.6 <–> WKT/PROJ.4 translation](http://cfconventions.org/wkt-proj-4.html)
* [Coordinate Systems Worldwide](https://epsg.io)

AliceBalfanz:
Alternative solution of On-the-fly reprojection by xcube serve implemented in https://github.com/dcs4cop/xcube/releases/tag/v0.11.0

PR #226, opened 2019-11-10
Forman 112 improved schema

forman:
Tiny step towards #112:

* Added missing tests for CubeSchema
* Added xarray accessor `xcube.schema` with HTML representation for NB


forman:
I merge now even without @dzelge 's approval. Need this for the DCFS demo and the next PR...

PR #230, opened 2019-11-18
write_cube to s3 bucket

AliceBalfanz:
This PR implements the upload of an xucbe dataset with write_cube into an s3 bucket. The client_kwargs (access key and access id) need to be specified in a dict. The path to the bucket is passed as an url, the same as it is in open_cube.


CHANGES.md
```diff
@@ -1,7 +1,8 @@
 ## Changes in 0.3.0 (in development)
 
 ### New
-
+* Added ability to upload xcube datasets in zarr format to s3 bucket within
```
forman: This is not necessarily an *upload*. Just say *write to*.

```suggestion
* Added ability to write xcube datasets in zarr format into s3 bucket using
```

CHANGES.md
```diff
@@ -1,7 +1,8 @@
 ## Changes in 0.3.0 (in development)
 
 ### New
-
+* Added ability to upload xcube datasets in zarr format to s3 bucket within 
+  xcube.core.dsio.write() #224
```
```suggestion
  xcube.core.dsio.write_cube() #224
```

test/webapi/test_s3buckethandlers.py
```diff
@@ -27,7 +27,7 @@ class S3BucketHandlersTest(unittest.TestCase):
 
     @unittest.skipUnless(XCUBE_SERVER_IS_RUNNING, SKIP_HELP)
     def test_open_cube_from_xube_server(self):
-        ds = open_cube('s3bucket/local', format_name='zarr', endpoint_url=SERVER_URL)
+        ds = open_cube('s3bucket/local', format_name='zarr', client_kwargs={'endpoint_url':SERVER_URL})
```
forman: This makes it harder for users to use `open_cube`. I'd prefer to keep `endpoint_url` and add it to `client_kwargs` internally.

AliceBalfanz:
Closing this PR because of further work on this branch. It will be reopened after implementing tests.

xcube/core/dsio.py
```diff
@@ -549,3 +553,28 @@ def rimraf(path):
...
             warnings.warn(f"failed to remove file {path}")
             pass
+
+
+def _get_path_or_store(path, client_kwargs, mode, root):
+    path_or_store = path
+    anon_mode = True
+
+    if client_kwargs is not None:
+        if 'aws_access_key_id' in client_kwargs and 'aws_secret_access_key' in client_kwargs:
```
dzelge: I think it would be great if you could attempt to get the credentials from system env variables when the user does not
provide them in the client_kwargs. That is safer for docker containers and the like.

e.g.:

```
# Get the value of 'HOME' 
# environment variable 
key = 'HOME'
value = os.getenv(key) 
```

AliceBalfanz:
Reopening after making changes based on above comments

AliceBalfanz:
What should I do with the conflict in the CHANGES.md? Because if I understand correct, I do want to merge it into the master branch and this implementation is planned for 0.4.0 release.

CHANGES.md
```diff
@@ -1,7 +1,12 @@
+## Changes in 0.4.0 (in development)
+
+### Enhancements 
+* Added ability to write xcube datasets in zarr format into s3 bucket using
+  xcube.core.dsio.write_cube() #224
```
```suggestion
  `xcube.core.dsio.write_cube()`. (#224)
```

xcube/core/dsio.py
```diff
@@ -549,3 +553,39 @@ def rimraf(path):
         except OSError:
             warnings.warn(f"failed to remove file {path}")
             pass
+
+
+def _get_path_or_store(path, client_kwargs, mode, root):
```
forman: Add types!

xcube/core/dsio.py
```diff
@@ -549,3 +553,39 @@ def rimraf(path):
...
+    path_or_store = path
+    anon_mode = True
+    if not client_kwargs:
+        client_kwargs = {}
+
+    aws_credentials = ['aws_access_key_id', 'aws_secret_access_key']
```
forman: Why the `aws_` prefix in the keyword argument names? How can these be valid for OTC and GCP object storage API as well?

xcube/core/dsio.py
```diff
@@ -418,31 +418,24 @@ def fitness(self, path: str, path_type: str = None) -> float:
...
-                    endpoint_url = f'{url.scheme}://{url.host}'
-                root = url.path
-                if root.startswith('/'):
-                    root = root[1:]
+                client_kwargs['region_name'] = kwargs.pop('region_name')
+
+            path_or_store, root, client_kwargs = _get_path_or_store(path_or_store, client_kwargs, mode, root)
 
-            if endpoint_url and root is not None:
+            if "endpoint_url" in client_kwargs and root is not None:
```
forman:
```suggestion
            if 'endpoint_url' in client_kwargs and root is not None:
```

xcube/core/dsio.py
```diff
@@ -418,31 +418,24 @@ def fitness(self, path: str, path_type: str = None) -> float:
...
+        mode = "read"
+        root = None
 
         if isinstance(path, str):
-            endpoint_url = None
-            region_name = None
-            root = None
-
+            client_kwargs = {}
+            if "client_kwargs" in kwargs:
```
forman:
```suggestion
            if 'client_kwargs' in kwargs:
```

xcube/core/dsio.py
```diff
@@ -418,31 +418,24 @@ def fitness(self, path: str, path_type: str = None) -> float:
     def read(self, path: str, **kwargs) -> xr.Dataset:
         path_or_store = path
         consolidated = False
+        mode = "read"
```
forman:
```suggestion
        mode = 'read'
```

environment.yml
```diff
@@ -14,6 +14,7 @@ dependencies:
   - fiona >=1.8
   - gdal >=2.4,<3.0
   - matplotlib >=3.0
+  - moto >=1.3
```
forman: Move down to "# Testing" section.

xcube/core/dsio.py
```diff
@@ -458,9 +451,20 @@ def write(self,
               output_path: str,
               compress=True,
               cname=None, clevel=None, shuffle=None, blocksize=None,
-              chunksizes=None):
+              chunksizes=None,
+              client_kwargs=None,
+              **kwargs):
+        mode = "write"
```
forman:
```suggestion
        mode = 'write'
```

xcube/core/dsio.py
```diff
@@ -458,9 +451,20 @@ def write(self,
...
+        root = None
+        path_or_store, root, client_kwargs = _get_path_or_store(output_path, client_kwargs, mode, root)
+
+        # keeping the code commented out below, because might be useful for makinging write able to change bucket policy
+        # if 'acl' in kwargs:
+        #     acl = kwargs.pop('acl')
         encoding = self._get_write_encodings(dataset, compress, cname, clevel, shuffle, blocksize, chunksizes)
-        dataset.to_zarr(output_path, mode="w", encoding=encoding)
+        dataset.to_zarr(path_or_store, mode="w", encoding=encoding)
```
forman:
```suggestion
        dataset.to_zarr(path_or_store, mode='w', encoding=encoding)
```

xcube/core/dsio.py
```diff
@@ -549,3 +553,39 @@ def rimraf(path):
         except OSError:
             warnings.warn(f"failed to remove file {path}")
```
forman:
```suggestion
            warnings.warn(f'failed to remove file {path}')
```

xcube/core/dsio.py
forman: I don't really agree, at least in this very case here, because users might want to write to different object storage providers with different credentials. For the CLI, this might make sense, Python API users can env vars on their own.

xcube/core/dsio.py
```diff
@@ -458,9 +451,20 @@ def write(self,
...
+        path_or_store, root, client_kwargs = _get_path_or_store(output_path, client_kwargs, mode, root)
+
+        # keeping the code commented out below, because might be useful for makinging write able to change bucket policy
+        # if 'acl' in kwargs:
+        #     acl = kwargs.pop('acl')
         encoding = self._get_write_encodings(dataset, compress, cname, clevel, shuffle, blocksize, chunksizes)
-        dataset.to_zarr(output_path, mode="w", encoding=encoding)
+        dataset.to_zarr(path_or_store, mode="w", encoding=encoding)
+        # if acl and client_kwargs and s3:
+        #     s3.chmod(path=root, acl=acl, kwargs=client_kwargs)
```
forman: Remove comment, unless you need to keep it. If so, say why.

xcube/core/dsio.py
```diff
@@ -458,9 +451,20 @@ def write(self,
...
+              chunksizes=None,
+              client_kwargs=None,
+              **kwargs):
+        mode = "write"
+        root = None
+        path_or_store, root, client_kwargs = _get_path_or_store(output_path, client_kwargs, mode, root)
+
+        # keeping the code commented out below, because might be useful for makinging write able to change bucket policy
+        # if 'acl' in kwargs:
+        #     acl = kwargs.pop('acl')
```
forman: Remove comment, unless you need to keep it. If so, say why.

forman:
@AliceBalfanz, please also resolve existing conflict

xcube/core/dsio.py
AliceBalfanz: It works for OTC as well, the provider is defined by the endpoint url. This is simply how the key is called to pass the credentials: https://s3fs.readthedocs.io/en/latest/#credentials

xcube/core/dsio.py
AliceBalfanz: Well, there is a comment describing, why it is kept just above :)

xcube/core/dsio.py
AliceBalfanz: I do not fully understand. Is this issue solved by the information, that this is just the key the s3fs library uses and that by defining the endpoint url it may also apply to OTC buckets? Or should I change something within the implementation? If so, I do need hints.

xcube/core/dsio.py
forman: But `aws` means Amazon Web Services. there this a wrong name for the key.  Please use a generic name such as `provider_acess_key_id` and `provider_access_key_secret`.

xcube/core/dsio.py
forman: Helge's suggestion is to use environment variables. But that prevents using different object storages at the same time, because environment variables are global.

forman:
Lets get this merged tomorrow!

dzelge:
Dann mal los! Supa!

Issue #206, opened 2019-10-16
xcube gen performs inaccurate reprojection (tbc)


forman:
From diverse investigations it turned out that the shift of cells as seen in https://github.com/dcs4cop/xcube-viewer/issues/64 may probably originate from innaccurate reprojection in `xcube gen` tool.

A first idea is to parameterise `xcube gen` to perform a reprojection with more ground control points so we perform a quantitative anaylsis.

AliceBalfanz:
One reason of the shift is that the reprojection of the data set is performed on the whole extent of the input file. Part avoiding the shift would be to first make a subset of the area and then perform the reprojection. Furthermore one xcube dataset should not cover a very large area. If an area of interest is large, the solution might be to produce several xcube datasets which then are combined via the server/viewer to appear as one data set.

forman:
Inaccuracy isn't a bug.

forman:
Fixed by new rectification

PR #235, opened 2019-11-19
spatial subset before reprojection

AliceBalfanz:
This PR relates to #206 . It allows xcube to make a spatial subset according to the output region before performing the reprojection of the dataset.

### DIFF
--- a/xcube/core/gen/iproc.py
+++ b/xcube/core/gen/iproc.py
@@ -120,6 +120,17 @@ class InputProcessor(ExtensionComponent, metaclass=ABCMeta):
         """
         return None
 
+    def get_spatial_subest(self, dataset: xr.Dataset, dst_region: Tuple[float, float, float, float]) -> xr.Dataset:
+        """
+        Get spatial subset based on dst_region.
+
+        Returns ``dataset`` by default.
+        :param dataset: The dataset.
+        :param dst_region: The region taken for the spatial subset
+        :return: A spatial subset of the dataset.
+        """
+        return dataset
+
     def pre_process(self, dataset: xr.Dataset) -> xr.Dataset:
         """
         Do any pre-processing before reprojection.
--- a/xcube/core/gen/iproc.py
+++ b/xcube/core/gen/iproc.py
@@ -120,7 +120,7 @@ class InputProcessor(ExtensionComponent, metaclass=ABCMeta):
         """
         return None
 
-    def get_spatial_subest(self, dataset: xr.Dataset, dst_region: Tuple[float, float, float, float]) -> xr.Dataset:
+    def get_spatial_subset(self, dataset: xr.Dataset, dst_region: Tuple[float, float, float, float]) -> xr.Dataset:
         """
         Get spatial subset based on dst_region.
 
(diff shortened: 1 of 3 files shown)


xcube/core/gen/iproc.py
```diff
@@ -260,6 +271,24 @@ def pre_process(self, dataset: xr.Dataset) -> xr.Dataset:
 
         return _normalize_lon_360(dataset)
 
+    def get_spatial_subset(self, dataset: xr.Dataset, dst_region: Tuple[float, float, float, float]) -> xr.Dataset:
+        lon_min, lat_min, lon_max, lat_max = dst_region
```
forman: First make a test, if `dataset` bounds are entirely included in `dst_region`. If so, return `dataset` unchanged:

```python
from xcube.core.geom import get_dataset_geometry
from xcube.core.geom import get_box_split_bounds_geometry

ds_bounds_geom = get_dataset_geometry(dataset)
dst_region_geom = get_box_split_bounds_geometry(*dst_region)
if dst_region_geom.contains(ds_bounds_geom):
    return dataset
```

xcube/core/gen/iproc.py
```diff
@@ -260,6 +271,24 @@ def pre_process(self, dataset: xr.Dataset) -> xr.Dataset:
...
+        dataset_subset = dataset.copy()
+        dataset_subset.coords['x'] = xr.DataArray(np.arange(0, dataset.lon.size), dims='lon')
+        dataset_subset.coords['y'] = xr.DataArray(np.arange(0, dataset.lat.size), dims='lat')
+        lon_subset = dataset_subset.lon.where((dataset_subset.lon >= lon_min) & (dataset_subset.lon <= lon_max),
+                                              drop=True)
+        lat_subset = dataset_subset.lat.where((dataset_subset.lat >= lat_min) & (dataset_subset.lat <= lat_max),
+                                              drop=True)
+        x1 = lon_subset.x[0]
```
forman: Similarily, make a test, if `dataset` bounds are entirely excluded from `dst_region`, in this case return None and let `gen_cube` ignore the input.

xcube/core/gen/iproc.py
```diff
@@ -120,6 +120,17 @@ def get_extra_vars(self, dataset: xr.Dataset) -> Optional[Collection[str]]:
         """
         return None
 
+    def get_spatial_subset(self, dataset: xr.Dataset, dst_region: Tuple[float, float, float, float]) -> xr.Dataset:
```
forman: Additional, important notes:

* `get_spatial_subset` should be abstract in `InputProcessor`.
* This `get_spatial_subset` implementation should go into `XYInputProcessor`.
* This `get_spatial_subset` implementation for `DefaultInputProcessor` is much simpler and much faster, because we rectified coordinates already.

forman: Maybe  `get_spatial_subset` should not be a new method at all `InputProcessor`. There are already too much - adding a lot of responsibility for developers to implement a new one.
We have already the `pre_process` method for pre-processing, and subsetting *is a* pre-processing.
It may be better if we move a `get_spatial_subset` into `xcube.core.geom` and use it in the default implementation of `XYInputProcessor`.
That way, `get_spatial_subset` could be implemented in the fastest possible way and work for datasets 2-D `lon`, `lat` coordinates with 1-D `x`, `y` coordinates  (such as Sentinel-2) and for WGS84 inputs that have 1-D  `lon`, `lat` coordinates.

xcube/core/gen/iproc.py
dzelge: Does ``` dataset_subset.lon```work when the coord_name is 'x'?

forman: Yes, and it expected to be 2-D.

AliceBalfanz:
Closing this PR for the Moment and will reopen after having implemented requested and discussed changes

Issue #231, opened 2019-11-19
Add a monoton coordinate test to xcube verify


dzelge:
**Describe the issue**

xcube verify accepts not monotonely increasing coordinates. It seems that xarray accepts 
marginally small deviations from monotony, 

**To Reproduce**
1. Go to jupyterhub and load the 0.08333 deg global cube
2. Perform a diff(dim='lon') on lon

**Expected behavior**

The verifier should not accept such a coordinate 

**Additional context**
xcube viewer shift when displaying ESDC cubes

forman:
Note, that latitude may also be monotonically *decreasing*.

forman:
Also note, that diff(lon) doesn't work if the dataset bounds intersect with the anti-meridian.

forman:
Removed "urgent" as current issue may also be fixed in data cube files, which is even better.

dzelge:
> 
> 
> Also note, that diff(lon) doesn't work if the dataset bounds intersect with the anti-meridian.

In that case I will get one lon diff that is negative, right?

forman:
Right!

PR #236, opened 2019-11-19
Verify equidistance of spatial coordinates 

dzelge:
When accepting this PR, xcube verify will check whether
spatial coordinates are equidistant. This includes their bound coordinates 
if they exist. (see #231)


xcube/core/verify.py
```diff
@@ -81,9 +81,38 @@ def verify_cube(dataset: xr.Dataset) -> List[str]:
...
+        if xy_var_names[0] + '_bnds' in dataset.variables:
+            _check_coord_equidistance(dataset, xy_var_names[1] + '_bnds', xy_var_names[1], report)
+
     return report
 
 
+def _check_coord_equidistance(dataset, coord_name, dim_name, report, rtol=1e-09):
+    coordinate = dataset[coord_name]
+
+    if 'bnds' in coord_name:
```
forman: Use function that determines the bounds variable of a coordinate variable.   
`coord_var_bnds_name = coord_var.attrs.get("bounds", coord_var_name + "_bnds")`

examples/serve/demo/cube-1-250-250.zarr/.zattrs
```diff
@@ -1,4 +1,4 @@
 {
     "Conventions": "CF-1.7",
-    "coordinates": "time_bnds lat_bnds lon_bnds"
+    "coordinates": "time_bnds lon_bnds lat_bnds"
```
forman: Strange, these should actually be listed in the order of their dimensions, which is 

```suggestion
    "coordinates": "time_bnds lat_bnds lon_bnds"
```

However, it doesn't seem to matter.

CHANGES.md
```diff
@@ -17,6 +17,8 @@
   hidden `xcube apply` command. (#167) 
 * Added new function `xcube.core.geom.rasterize_features()` 
   to rasterize vector-data features into a dataset. (#222)
+* Extended verify to check whether spatial coordinates and their bounds are equidistant. (#231)
```
```suggestion
* Extended CLI command `xcube verify` and API function `xcube.core.verify.verify_cube` to check whether spatial coordinate variables and their associated bounds variables are equidistant. (#231)
```

CHANGES.md
```diff
@@ -89,6 +91,8 @@ versions.
 
 * Reorganisation of the Documentation and Examples Section (partly addressing #106)
 * Loosened python conda environment to satisfy conda-forge requirements
+* xcube is now deployed on conda-forge
```
```suggestion
* xcube is now available as a conda package on the conda-forge channel. To install latest xcube package, you can now type: `conda install -c conda-forge xcube`
```

CHANGES.md
```diff
@@ -89,6 +91,8 @@ versions.
...
+* Changed the unittesting code to minimize 3rd party warning
```
```suggestion
* Changed the unittesting code to minimize warnings reported by 3rd-party packages
```

xcube/cli/verify.py
```diff
@@ -40,6 +40,7 @@ def verify(cube):
       variables, if any;
     * has any data variables and that they are valid, e.g. min. 3-D, all have
       same dimensions, have at least dimensions "time", "lat", "lon".
+    * spatial coordinates and their corresponding bounds (if exist) are equidistant
```
forman: Isn't it now:

```suggestion
    * spatial coordinates and their corresponding bounds (if exist) are equidistant
       and monotonically increasing or decreasing.
```

?

xcube/core/verify.py
```diff
@@ -81,9 +81,41 @@ def verify_cube(dataset: xr.Dataset) -> List[str]:
...
+            report.append(f"Coordinate {bnds_name} is not equidistant")
+        elif not _check_equidistance_from_diff(diff[:, 1], rtol=rtol):
+            report.append(f"Coordinate {bnds_name} is not equidistant")
+
+
+def _check_equidistance_from_diff(diff, rtol=None):
+    if rtol is None:
+        rtol = np.abs(np.divide(diff[0], 100.00)).values
+
+    # Check whether teh bounding box intersect with the anti-meridian.
```
forman:
```suggestion
    # Check whether the bounding box intersects with the anti-meridian.
```

xcube/core/verify.py
```diff
@@ -81,9 +81,41 @@ def verify_cube(dataset: xr.Dataset) -> List[str]:
...
+    ct_neg = diff.where(diff < 0).count().values
+
+    if ct_neg == 1:
+        return np.allclose(diff[0], diff.where(diff > 0, drop=True), rtol=rtol)
```
forman:
```suggestion
        # <please explain, why you use diff.where(diff > 0, drop=True) here>
        return np.allclose(diff[0], diff.where(diff > 0, drop=True), rtol=rtol)
```

xcube/core/verify.py
```diff
@@ -81,9 +81,41 @@ def verify_cube(dataset: xr.Dataset) -> List[str]:
...
+        _check_coord_equidistance(dataset, xy_var_names[0], xy_var_names[0], report)
+        _check_coord_equidistance(dataset, xy_var_names[1], xy_var_names[1], report)
+
     return report
 
 
+def _check_coord_equidistance(dataset, coord_name, dim_name, report, rtol=None):
+    diff = dataset[coord_name].diff(dim=dim_name)
+    if not _check_equidistance_from_diff(diff, rtol=rtol):
+        report.append(f"Coordinate {coord_name} is not equidistant")
```
forman: Always try being consistent! Here be consistent with other report strings in this module. Therefore

```suggestion
        report.append(f"coordinate variable {coord_name!r} is not equidistant")
```

xcube/core/verify.py
```diff
@@ -81,9 +81,41 @@ def verify_cube(dataset: xr.Dataset) -> List[str]:
...
+
+    bnds_name = dataset.attrs.get('bounds', f'{coord_name}_bnds')
+
+    if bnds_name in dataset.coords:
+        diff = dataset[bnds_name].diff(dim=dim_name)
+        if not _check_equidistance_from_diff(diff[:, 0], rtol=rtol):
+            report.append(f"Coordinate {bnds_name} is not equidistant")
```
forman:
```suggestion
            report.append(f"coordinate variable {bnds_name!r} is not equidistant")
```

xcube/core/verify.py
```diff
@@ -81,9 +81,41 @@ def verify_cube(dataset: xr.Dataset) -> List[str]:
...
+        elif not _check_equidistance_from_diff(diff[:, 1], rtol=rtol):
+            report.append(f"Coordinate {bnds_name} is not equidistant")
```
forman:
```suggestion
            report.append(f"coordinate variable {bnds_name!r} is not equidistant")
```

xcube/core/verify.py
dzelge: Short question: At this stage, the verification excludes the anti-medirian intersecting bound from the verification. Is that OK? We could compute the distance by adding 360degrees to the difference. But that would be projection specific.

forman: Good point. We should at least detect anti-meridian crossing, and if we cross, skip equidistance checking for time being (will implement for 0.4). Use `is_lon_lat_dataset()` to distinguish from other CRS. 
In various places, I use the condition `lon_start > lon_end` to detect anti-meridian crossing for bboxes.

forman: see
* https://github.com/dcs4cop/xcube/blob/8ff6fcbe73a2145de48958d541c283d9583009f2/xcube/webapi/controllers/tiles.py#L262
* https://github.com/dcs4cop/xcube/blob/8ff6fcbe73a2145de48958d541c283d9583009f2/xcube/core/geom.py#L453
* https://github.com/dcs4cop/xcube/blob/520cd2cf7e3f4d2d121889064bdf3732863c244f/xcube/util/tilegrid.py#L214

Issue #246, opened 2019-11-25
Presorting input list fails xcube gen when input data is corrupt


AliceBalfanz:
**Describe the bug**
When there is a corrupt input file in the list of input files for the xcube gen process, the presorting function does not skip the input file, but the whole process is failing: 

```
 xcube gen -c ~/projects/xcube-services/dcs4cop/xcube-gen-configs/dcs4cop-gen_BC_config_S2_UseCase_UHULL_31UCV.yml ../VITO_S2/0025_UHULL/*31UCV*.nc
/home/xcube/projects/xcube/xcube/core/gen/gen.py:92: UserWarning: append_mode in gen_cube() is deprecated, time slices will now always be inserted, replaced, or appended.
  warnings.warn('append_mode in gen_cube() is deprecated, '
OS error: [Errno -101] NetCDF: HDF error: b'/mnt/dcs4cop/VITO_S2/0025_UHULL/DCS4COP_DEFAULT_S2A_0025_UHULL_20190817T105621Z_31UCV_MPS_V100.nc'
Traceback (most recent call last):
  File "/home/xcube/miniconda3/envs/xcube/lib/python3.7/site-packages/xarray/backends/file_manager.py", line 198, in _acquire_with_cache_info
    file = self._cache[self._key]
  File "/home/xcube/miniconda3/envs/xcube/lib/python3.7/site-packages/xarray/backends/lru_cache.py", line 53, in __getitem__
    value = self._cache[key]
KeyError: [<class 'netCDF4._netCDF4.Dataset'>, ('/mnt/dcs4cop/VITO_S2/0025_UHULL/DCS4COP_DEFAULT_S2A_0025_UHULL_20190817T105621Z_31UCV_MPS_V100.nc',), 'r', (('clobber', True), ('diskless', False), ('format', 'NETCDF4'), ('persist', False))]

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/xcube/projects/xcube/xcube/cli/main.py", line 52, in main
    exit_code = cli.main(args=args, obj=ctx_obj, standalone_mode=False)
  File "/home/xcube/miniconda3/envs/xcube/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/xcube/miniconda3/envs/xcube/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
[...]

PR #247, opened 2019-11-25
skip corrupt input file

AliceBalfanz:
This PR makes xcube gen skip a corrupt input file when presorting the input files to a list. Currently the process is failing.  
The issue is described in #246


CHANGES.md
```diff
@@ -1,5 +1,10 @@
 ## Changes in 0.4.0 (in development)
```
forman: Remove this line.

xcube/core/gen/gen.py
```diff
@@ -352,10 +352,15 @@ def _get_sorted_input_paths(input_processor, input_paths: Sequence[str]):
...
-            input_path_list.append(input_file)
-            tuple_seq = zip(time_list, input_path_list)
+        try:
+            with xr.open_dataset(input_file) as dataset:
+                t1, t2 = input_processor.get_time_range(dataset)
+                time_list.append((t1 + t2) / 2)
+                input_path_list.append(input_file)
+                tuple_seq = zip(time_list, input_path_list)
+        except OSError:
+            traceback.print_exc()
```
forman: Output warning saying that input has been skipped. Use `monitor` from outer scope.

forman: Don't print entire stack trace here, this wont help any user, just report the OSError's message.

CHANGES.md
```diff
@@ -1,5 +1,10 @@
...
 
+## Changes in 0.3.1 (in development)
+
+### Fixes
+* `xcube gen` does not fail when a corrupt input file is among the input files. #246
```
forman: The implementation does not check for *corrupt* files. Instead, it skips *any* input it can't open. Say so here or change implementation.

PR #283, opened 2020-03-04
use chunksizes as tile sizes for rectification

AliceBalfanz:
This PR implements the usage within `xcube gen` of chunksizes as tile sizes for rectification if they are given via the `output_writer_params` and contain at least 'lat' or 'lon'.


CHANGES.md
```diff
@@ -42,7 +42,8 @@
   reprojection method in `xcube.core.gen.iproc.XYInputProcessor` and
   `xcube.core.gen.iproc.DefaultInputProcessor`, if ground control points are not 
   specified, i.e. the input processor is configured with `xy_gcp_step=None`. (#206)
-
+* Tile sizes for rectification in `xcube gen`are now derived from output_writer_params if given and
```
```suggestion
* Tile sizes for rectification in `xcube gen` are now derived from `output_writer_params` if given in configuration and 
```

CHANGES.md
```diff
@@ -42,7 +42,8 @@
   reprojection method in `xcube.core.gen.iproc.XYInputProcessor` and
   `xcube.core.gen.iproc.DefaultInputProcessor`, if ground control points are not 
   specified, i.e. the input processor is configured with `xy_gcp_step=None`. (#206)
-
+* Tile sizes for rectification in `xcube gen`are now derived from output_writer_params if given and 
+  if it contains chunksizes for 'lat' or 'lon', in order to use dask.
```
```suggestion
  if it contains a `chunksizes` parameter for 'lat' or 'lon'. This will force the generation of a chunked xcube dataset and will utilize Dask arrays for out-of-core computations. This is very useful for large data cubes whose time slices would otherwise not fit into memory.
```

xcube/core/gen/iproc.py
```diff
@@ -328,7 +328,9 @@ def process(self,
                                       geo_coding=geo_coding,
                                       output_geom=output_geom,
                                       is_y_reversed=True)
-
+            if output_geom.is_tiled:
+                if dataset.chunks.get('lat')[0] < dataset.chunks.get('lat')[-1]:
```
forman:
```suggestion
                # The following condition may become true, if we have used rectified_dataset(input, ..., is_y_reverse=True)
                # In this case y-chunksizes will also be reversed. So that the first chunk is smaller than any other.
                # Zarr will reject such datasets, when written.
                if dataset.chunks.get('lat')[0] < dataset.chunks.get('lat')[-1]:
```

PR #284, opened 2020-03-10
Additional timer for time series

AliceBalfanz:
This PR adds more timers for racing time of retrieving time series and made a change to logging output, so that more logs can be associated to a particular dataset via the dataset identifier. No changes in code structure was made.


AliceBalfanz:
Closing because of further changes, which I am going to make.

PR #285, opened 2020-03-13
Fix dockerfile such that serving from docker container works

RmStorm:
When serving xcubes from inside a docker container the address has to be
set to 0.0.0.0 for it to work. This is not documented anywhere and the
error message was not quite clear. I've added a docker compose file so
all anyone has to do is 'docker-compose up -d' this will spin up the
image in such a way that the xcube-viewer immediately picks up on it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ FROM continuumio/miniconda3:latest
 MAINTAINER helge.dzierzon@brockmann-consult.de
 
 LABEL name=xcube
-LABEL version=0.3.0.dev1
+LABEL version=0.4.0.dev0
 LABEL conda_env=xcube
 
 # Ensure usage of bash (simplifies source activate calls)
@@ -42,4 +42,4 @@ EXPOSE 8000
 
 # Start server
 ENTRYPOINT ["/bin/bash", "-c"]
-CMD ["source activate xcube && xcube"]
+CMD ["source activate xcube && xcube serve --address 0.0.0.0 --port 8000 --config ./examples/serve/demo/config.yml --verbose"]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,7 @@
+version: '3'
+
+services:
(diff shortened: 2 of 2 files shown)


RmStorm:
Fyi, I'm a software developer at [NIVA](https://github.com/NIVANorge) and I want to deploy the viewer and xcube server in our cloud infrastructure on google cloud. For now I will just package the xcubes in the docker container. In the future I will try to store them in the google blob object storage which is very similar to Amazon's. At that time I might make one or two additional PR's.

RmStorm:
Hi @dzelge,

Thanks, I had found that repo but it didn't quite work for me. I made a docker compose file that pulls the quay.io xcube image like this:
```
version: '3'
services:
  example-server:
    entrypoint: ""
    command: "source activate xcube && xcube serve -n example --prefix 'api/v1.dev' -p 8000 -v -u 30 -c /example/server-config.yml --traceperf"
    image: quay.io/dzelge/xcube:latest
    restart: always
    ports:
      - "0.0.0.0:8080:8000"
    volumes:
      - .:/example
```
When I run that I get this error:
`ERROR: for example-server  Cannot start service example-server: OCI runtime create failed: container_linux.go:349: starting container process caused "exec: \"source\": executable file not found in $PATH": unknown`

I don't think this PR/repo is a good place to discuss issues with https://github.com/dcs4cop/xcube-docker but I just thought I'd let you know.

Because the above didn't work I thought the image in this repository was more promising which is how we ended up here :smiley:. We don't use docker-compose for production either but it's extremely convenient to run a local copy of the deployed cloud infrastructure (our new cloud infrastructure is pretty much completely docker based).

So, it's up to you if you want to merge this PR or if the dockerfile in this repo will be removed. If you decide on the second option I will figure out how to locally run infrastructure using the images from the other repo and make a PR there with updated documentation/docker-compose file.

dzelge:
Obsolete. Closing this.

PR #310, opened 2020-06-02
Use mamba in travis and docker builds

### LINKED ISSUE #308: Try replacing conda with mamba for Travis and Docker builds
conda's "solving environment" process can be very slow when creating an environment. For Travis xcube builds it can exceed ten minutes. Since conda produces no output while solving the environment, this causes an automatic time-out (see e.g. [here](https://travis-ci.com/github/dcs4cop/xcube/jobs/340922052)). Possible solutions:

1. Add `-v -v` flags with `conda env create` to enable debug-level verbosity, which produces output while the solver is running. Undesirable because the overwhelming amount of debug output would make it harder to debug build problems.

2. Use Travis' [`travis_wait`](https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received) function, which was designed for cases like this. It wraps another command while writing a line to the build log every minute to defeat the time-out.

3. Use [mamba](https://github.com/QuantStack/mamba), a recently released drop-in replacement for conda, claimed to be "several orders of magnitude faster than conda to resolve compatibility contraints of conda environment". It currently carries the warning **"Still in BETA! Don't use mamba to install something in valuable environments!"**.

Provided that mamba does work as expected, option 3 seems the most attractive.

pont-us:
Fixes #308 .

This branch modifies the Travis build script to install the environment using [mamba](https://github.com/QuantStack/mamba) rather than conda, which vastly reduces build times and prevents Travis from killing a build due to a time-out during conda's lengthy "solving environment" process. Two further changes were required to make this work:

 - A minor syntax change in the `environment.yml` file to make it compatible with mamba
 - A change to the `get_projection_wkt` function in `xcube.core.reproject` to make its output standard-compliant; GDAL 2.4 (as selected when installing the environment with conda) didn't mind the missing `UNIT` element, but mamba on mac os selects GDAL 3 instead, which throws an error.


pont-us:
> Good stuff. Can you please adapt the Dockerfile as well please?
> H

Done!

PR #313, opened 2020-06-03
ImageGeom.from_dataset height check

### LINKED ISSUE #303: ImageGeom.from_dataset(new_cube()) gives "ValueError: invalid  y_min"
**Description**
If a datacube created from `xcube.core.new.new_cube()` (with no arguments specified) is passed to `xcube.core.imgeom.ImageGeom.from_dataset` as the only argument, the latter throws a `ValueError: invalid y_min`.

**Steps to reproduce**
```
from xcube.core.new import new_cube
from xcube.core.imgeom import ImageGeom
image_geom = ImageGeom.from_dataset(new_cube())
```

**Result**
```
Traceback (most recent call last):
  File "bug.py", line 3, in <module>
    image_geom = ImageGeom.from_dataset(new_cube())
  File "/home/pont/loc/xcube-repo/xcube/xcube/core/imgeom.py", line 246, in from_dataset
    ij_denom=ij_denom)
  File "/home/pont/loc/xcube-repo/xcube/xcube/core/imgeom.py", line 284, in _compute_output_geom
    is_geo_crs=geo_coding.is_geo_crs)
  File "/home/pont/loc/xcube-repo/xcube/xcube/core/imgeom.py", line 72, in __init__
    raise ValueError('invalid y_min')
ValueError: invalid y_min
```

**Excepted behaviour**
`ImageGeom.from_dataset(new_cube())` returns a valid `ImageGeom` instance rather than throwing an exception.

pont-us:
When creating an ImageGeom from a dataset, correct the height if it would otherwise give a maximum latitude >90°. Fixes #303.


xcube/core/imgeom.py
```diff
@@ -277,6 +277,11 @@ def _compute_output_geom(dataset: xr.Dataset,
     dst_height = 1 + math.floor((src_y_max - src_y_min) / src_xy_res)
```
forman, line 277:
```suggestion
    dst_height = math.round((src_y_max - src_y_min) / src_xy_res) or 1
```

xcube/core/imgeom.py
```diff
@@ -277,6 +277,11 @@ def _compute_output_geom(dataset: xr.Dataset,
...
     dst_width = i_denom * ((dst_width + i_denom - 1) // i_denom)
     dst_height = j_denom * ((dst_height + j_denom - 1) // j_denom)
+
+    # Reduce height if it takes the maximum latitude over 90° (see Issue #303).
+    if src_y_min + dst_height * src_xy_res > 90.0:
+        dst_height -= 1
+
```
forman, line 284: What about `dst_width`?

xcube/core/imgeom.py
pont-us: See [my comment](https://github.com/dcs4cop/xcube/issues/303#issuecomment-638114126) on the issue. The preferred approach depends on the desired behaviour: `1 + math.floor...` will always provide enough pixels to cover the desired area. `math.round` will sometimes round down so that not all the desired area is covered -- but since the missing cropped area is less than a pixel-width, this is probably just nitpicking.

xcube/core/imgeom.py
pont-us: Copy-pasting form [my comment](https://github.com/dcs4cop/xcube/issues/303#issuecomment-638114126) on the issue:

> In the case of the `new.new_cube()` output, it happens that the width does not exceed 360°. In this case `359 / xy_res` =~ 507.70, so the rounding behaviour doesn't add enough to take `width * xy_res` over 360°. Even tweaking the oversampling value passed to `ImageGeom.from_dataset` doesn't seem to put the width over 360°; the longitudinal check has slightly more leeway since it's looking directly at the total width, whereas the latitudinal check is adding the height to `y_min`, which in this case is already offset by 0.5° from the minimum possible value of -90°. So it's probably not worth adding an explicit width correction, unless it later becomes clear that the width can indeed somehow exceed 360°.

forman: Ok!

xcube/core/imgeom.py
forman: Ok, agree!

pont-us:
> Please document change in `CHANGES.md`!

Done!

PR #320, opened 2020-06-23
Implemented new S3 store

forman:
In this branch, I

- implemented a new S3 store;
- added and improved tests for other store implementations.

### DIFF
--- a/xcube/core/store/accessors/dataset.py
+++ b/xcube/core/store/accessors/dataset.py
@@ -23,12 +23,13 @@ from typing import Dict, Any, Tuple, Optional
 import s3fs
 import xarray as xr
 
+from xcube.core.store.accessor import DataAccessorError
 from xcube.core.store.accessor import DataOpener
 from xcube.core.store.accessor import DataWriter
-from xcube.core.store.accessor import DataAccessorError
 from xcube.core.store.accessors.posix import PosixDataDeleterMixin
 from xcube.util.assertions import assert_instance
-from xcube.util.jsonschema import JsonBooleanSchema, JsonArraySchema
+from xcube.util.jsonschema import JsonArraySchema
+from xcube.util.jsonschema import JsonBooleanSchema
 from xcube.util.jsonschema import JsonObjectSchema
 from xcube.util.jsonschema import JsonStringSchema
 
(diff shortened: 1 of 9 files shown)


xcube/core/store/accessors/dataset.py
```diff
@@ -18,20 +18,22 @@
...
 import xarray as xr
 
 from xcube.core.store.accessor import DataOpener
 from xcube.core.store.accessor import DataWriter
-from xcube.core.store.accessors.posix import PosixDataDeleter
+from xcube.core.store.accessor import DataAccessorError
+from xcube.core.store.accessors.posix import PosixDataDeleterMixin
 from xcube.util.assertions import assert_instance
-from xcube.util.jsonschema import JsonBooleanSchema
+from xcube.util.jsonschema import JsonBooleanSchema, JsonArraySchema
```
TonioF, line 31: Split into two lines.

xcube/core/store/accessors/dataset.py
```diff
@@ -52,20 +54,60 @@ def write_data(self, data: xr.Dataset, data_id: str, replace=False, **write_para
...
+                    additional_properties=True,
+                ),
+                decode_cf=JsonBooleanSchema(
+                    description='Whether to decode these variables, assuming they were saved '
+                                'according to CF conventions.',
+                    default=True,
+                ),
+                mask_and_scale=JsonBooleanSchema(
+                    description='If True, replace array values equal to attribute "_FillValue" with NaN. '
+                                'Use attribute "scaling_factor" and "add_offset" to computed actual values.',
```
TonioF, line 80: What happens otherwise?

xcube/core/store/stores/memory.py
```diff
@@ -76,7 +79,7 @@ def search_data(self, type_id: str = None, **search_params) -> Iterator[DataDesc
                 yield new_data_descriptor(data_id, data)
 
     def get_data_opener_ids(self, data_id: str = None, type_id: str = None) -> Tuple[str, ...]:
-        return '*:*:memory',
+        return _ACCESSOR_ID,
```
TonioF, line 82: Looking at the the docstring, this should not return a pattern but a list of ids. Same is valid for get_data_writer_ids below.

xcube/core/store/accessors/dataset.py
forman: Just negate what is said there. But that's really redundant:

'If False, do nit replace array values equal to attribute "_FillValue" with NaN. '
                                'Do not use attributes "scaling_factor" and "add_offset" to compute actual values.'

xcube/core/store/stores/memory.py
forman: This has been in there since the beginning, and we talked about it, see the predicate factory function. Just using a constant now.

xcube/core/store/accessors/dataset.py
forman: Ok.

xcube/core/store/accessors/dataset.py
forman: Fixed some typos in description.

PR #321, opened 2020-06-23
Forman 307 storeframework cli

forman:
This PR adds a new, experimental, useful, and still hidden CLI tool `xcube io` to xcube:

```
$ xcube io --help
Usage: xcube io [OPTIONS] COMMAND [ARGS]...

  Tools for xcube's generic I/O system.

Options:
  --help  Show this message and exit.

Commands:
  opener  Tools for xcube's data openers.
  store   Tools for xcube's data stores.
  writer  Tools for xcube's data writers.
```


forman:
> I would very much welcome a command that lists the data sets a store can offer

That is possible:

    xcube io store info directory -D

AliceBalfanz:
Why are no parameters required for opening data from s3-buckets? I would expect, that somehow access id and key need to be passed in order to open a dataset. 
```
xcube io opener info dataset:zarr:s3`
xarray.Dataset in Zarr format stored in S3-like object storage
No parameters required.
```

AliceBalfanz:
Besides my question above, everything I tried worked nicely :)

pont-us:
It would be nice to have a friendlier error message for this case:

```
$ xcube io opener info does:not:exist
Internal error: 'NoneType' object has no attribute 'metadata'
```

forman:
> Why are no parameters required for opening data from s3-buckets? I would expect, that somehow access id and key need to be passed in order to open a dataset.
> 
> ```
> xcube io opener info dataset:zarr:s3`
> xarray.Dataset in Zarr format stored in S3-like object storage
> No parameters required.
> ```

If no credentials are passed, it switches to anonymous mode. However, it then actually go the "boto3-way" as discussed yesterday. Haven't had the time to look into that. Maybe you could?

forman:
> It would be nice to have a friendlier error message for this case:

I disagree. This is an internal error, usually caused by developers, not the CLI users.

PR #332, opened 2020-09-01
Added example for store instance configuration

forman:
This PR just adds an example for  store instance configuration for `xcube gen2`. (Made the changes long time ago!)


examples/gen2/configs/stores.json
```diff
@@ -0,0 +1,8 @@
+{
+  "local_examples_gen2": {
+    "store_id": "directory",
+    "store_params": {
+      "base_dir": "examples/gen2"
+    }
+  }
+}
```
dzelge, line 8:
```suggestion 
}

```

PR #333, opened 2020-09-08
Fix failing unit tests

AliceBalfanz:
Restricted package dependencies due to false behavior in newer version caused by unfixed bugs: 
- pandas >=0.24,<1.1
- s3fs >=0.2,<0.5


AliceBalfanz:
Will reopen PR after aplying more changes, and adapting code to suite dependencies version increases.

PR #336, opened 2020-09-21
Add date limits to JSON schema and document opener parameters API

pont-us:
- Add `min_datetime` and `max_datetime` to `JSONStringSchema` to make results of `get_open_data_params_schema` more fully descriptive.
 - Add `opener_parameters_api.md` to documentation (a detailed description of how the opener parameters JSON Schema should be used by openers and UI generators).


pont-us:
Travis builds are failing, but this is because of seven unit tests which are also currently failing in master.

dzelge (changes requested):
Cool docs. If I understand correctly time periods are specified only by a duration character e.g. M or Y. However, they are in the same chapter of the date/time RFC you quoted. I'd just mention it In the chapter Date, time, and duration. (I left a remark)

docs/source/opener_parameters_api.md
```diff
@@ -0,0 +1,220 @@
...
+   available variables.
+ - If a list containing *some* of the possible members is given, a dataset
+   corresponding to those members only MUST be returned. In the case of
+   `variables`, this will result in a dataset containing only the requested
+   variables.
+ - As a special case of the above, if an empty list is supplied, a dataset
+   containing *no data* MUST be returned -- but with the requested spatial and
+   temporal dimensions.
+
+## Date, time, and duration specifications
```
dzelge, line 116: If I understand correctly time periods are specified only by a duration character e.g. M or Y. However, they are in the same chapter of the date/time RFC you quoted.

docs/source/opener_parameters_api.md
```diff
@@ -0,0 +1,220 @@
...
+defined below, in order to keep the interface consistent. For instance, if a
+dataset allows a time aggregation period to be specified, it MUST use the
+`time_period` parameter with the format described below rather than some other
+alternative name and/or format. The parameters are described below with their
+Python type annotations.
+
+ - `variable_names: List[str]`  
+   A list of the identifiers of the requested variables.
+ - `bbox: Union[str,Tuple[float, float, float, float]]`
+   The bounding box for the requested data, in the order x0, y0, x1, y1.
```
forman, line 84:
```suggestion
   The bounding box for the requested data, in the order xmin, ymin, xmax, ymax. 
   Must be given in the units of the specified spatial coordinate reference system `crs`. 
```

docs/source/opener_parameters_api.md
```diff
@@ -0,0 +1,220 @@
...
+ - `crs: str`  
+   The identifier for the co-ordinate reference system of geographic data.
```
forman, line 86:
```suggestion
   The identifier for the spatial coordinate reference system of geographic data.
```

docs/source/opener_parameters_api.md
```diff
@@ -0,0 +1,220 @@
...
+ - `spatial_res: float`  
+   The requested spatial resolution (x and y) of the returned data.
```
forman, line 88:
```suggestion
   The requested spatial resolution (x and y) of the returned data.
   Must be given in the units of the specified spatial coordinate reference system `crs`. 
```

docs/source/opener_parameters_api.md
```diff
@@ -0,0 +1,220 @@
...
+ - `time_range: Tuple[Optional[str], Optional[str]]`  
+   The requested time range for the data to be returned. See section
+   ‘Date, time, and duration specifications’ below.
+ - `time_period: str`  
+   The requested temporal aggregation period for the data. See section
+   ‘Date, time, and duration specifications’ below.
```
forman, line 94: Is it possible to add in-page ref for _‘Date, time, and duration specifications’_?

docs/source/opener_parameters_api.md
```diff
@@ -0,0 +1,220 @@
...
+JSON Schema itself does not offer a way to impose time limits on a string
+schema with the `date` or `date-time` format. This is a problem for xcube
+generator UI creation, since it might be reasonably expected that a UI will
+show and enforce such limits. The xcube opener API therefore defines an
+unofficial extension to the JSON string schema: a `JsonStringSchema` object
+(as returned as part of a `JsonSchema` by a call to
+`get_open_data_params_schema`) MAY, if it has a `format` property with a value
+of `date` or `date-time`, also have one or both of the properties
+`min_datetime` and `max_datetime`. These properties must also conform to the
+`date` or `date-time` format.
```
forman, line 151:
```suggestion
`date` or `date-time` format.
xcube provides a dedicated `JsonDatetimeSchema` for this purpose. Internally, it extends
the `JsonStringSchema` by adding the required properties to the JSON string schema.
```

@pont-us Please add new class `JsonDatetimeSchema` and test. This will make code more readable!

docs/source/opener_parameters_api.md
```diff
@@ -0,0 +1,220 @@
...
+the `default` annotation and the UI generator SHOULD set the default value in
+the UI accordingly. If the `title` annotation is absent, the UI generator
+SHOULD use the key corresponding to the parameter's schema in the parent
+schema as a fallback.
+
+### Generalized conversion of parameter schemas
+
+For parameters other than the common parameters, the UI can be generated
+automatically from the schema structure. In the case of a GUI, a one-to-one
+conversion of instances of `JsonSchema` subclasses into GUI elements will
```
forman, line 204: I'd prefer a translation of _JSON schema_ to GUI elements here, rather than a translation of `JsonSchema` subclasses to GUI elements. In the GUI, we will receive plain JSON and we should make no assumption of how this JSON was generated.

xcube/util/jsonschema.py
```diff
@@ -46,6 +46,7 @@ def __init__(self,
                  title: str = None,
                  description: str = None,
                  examples: str = None,
+                 any_of: Sequence['JsonSchema'] = None,
```
forman, line 49: I have excluded the `anyOf`, `oneOf`, `allOf` keywords (https://json-schema.org/understanding-json-schema/reference/combining.html) from xcube by intention. They add complexity into the `from_instance` and `to_instance` method implementations. As far as I can see, you ignore the new `any_of` parameter in those implementations, which is not ok.

forman (changes requested):
Very goodm thanks a lot. However, as you can see, we should further discuss some points..

docs/source/opener_parameters_api.md
forman: Right!

pont-us: I deliberately left this vague, and should have marked it explicitly as a TODO. We should discuss exactly what we want to support as a period specification -- I couldn't find it detailed anywhere in the xcube docs.

docs/source/opener_parameters_api.md
```diff
@@ -0,0 +1,220 @@
...
+     opener parameters.
+
+  2. The schema does not allow the representation of interdependencies
+     between values of opener parameters within a dataset. For
+     instance, the following cannot be represented in an open
+     parameters schema:
+     
+     *sensor_type*: A or B  
+     *variables*: [temperature, humidity] for sensor type A;
+                  [temperature, pressure] for sensor type B
```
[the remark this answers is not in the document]
pont-us: Yes. I have now updated the document to clarify this.

docs/source/opener_parameters_api.md
```diff
@@ -0,0 +1,220 @@
...
+
+To work around some of the restrictions of point (2) above, a dataset MAY be
+presented by the opener as multiple "virtual" datasets with difference
+parameter schemas. For instance, the hypothetical dataset described above
+could be offered not as a single dataset `envdata` but as two datasets
+`envdata:sensor-a` and `envdata:sensor-b`, with different sets of permitted
```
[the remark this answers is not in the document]
pont-us: Done.

docs/source/opener_parameters_api.md
pont-us: Done.

xcube/util/jsonschema.py
pont-us: > As far as I can see, you ignore the new `any_of` parameter in those implementations, which is not ok.

Yes, that was an oversight. How should I proceed? Add support for `any_of` to `from_instance` and `to_instance`, or remove `any_of` entirely? And if the latter, what to do about date specifications?

1. Only accept `date-time` format?
2. Only accept `date` format?
3. Deactivate JSON Schema format validation and handle it elsewhere?
4. Implement alternative formats for the date specifiers using some other JSON Schema mechanism that doesn't require `any_of`? But I don't know of any such mechanism (that's why I used `any_of`).

forman: Do you know if `date-time` includes `date`? I.e. is "2020-09-23" a valid datetime. In Pandas it is not. In xarray sel() "2020-09-23" converts into a datetime range "2020-09-23T00:00:00.000" to "2020-09-23T23:59:59.999".

docs/source/opener_parameters_api.md
pont-us: Agreed, and updated accordingly.

docs/source/opener_parameters_api.md
pont-us: As to the idea that month units are forbidden for time periods in pandas or xarray, I can't find any evidence of this. For instance, [here's](http://xarray.pydata.org/en/stable/examples/weather-data.html?highlight=month#Monthly-averaging) an example of monthly averaging with xarray and [here's](https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#timeseries-offset-aliases) the extensive list of pandas time series frequency specifiers (which includes monthly). I'm not sure whether we'd want to explicitly support the powerful and complex pandas schema; perhaps a small subset would be safer.

pont-us: I suggest that, for now, we define the time period format as [integer][unit], where [unit] is one of HDWMY (hour, day, week, month, year), and [integer] may be omitted to indicate 1. This could even be validated by a simple JSON Schema format regex.

CHANGES.md
```diff
@@ -2,6 +2,10 @@
 
 * `normalize_dataset` now ensures that latitudes are decreasing.
 
+* Add documentation for the use of the opener parameters.
```
forman, line 5:
```suggestion
* Add documentation for the use of the open parameters passed to `xcube.core.store.DataOpener.open_data()`.
```

CHANGES.md
```diff
@@ -2,6 +2,10 @@
...
+
+* Various improvements to the JSON Schema classes.
```
forman, line 7:
```suggestion
* The JSON Schema classes in `xcube.util.jsonschema` have been extended in various way:
  - ...
  - ...
```

xcube/util/jsonschema.py
```diff
@@ -158,6 +168,51 @@ def to_dict(self) -> Dict[str, Any]:
         return d
 
 
+class JsonDatetimeSchema(JsonStringSchema):
+    # noinspection PyShadowingBuiltins
+    def __init__(self,
+                 min_datetime: str = None,
+                 max_datetime: str = None,
+                 format='date-time',
```
forman, line 176: If `format` can also be `"date"` then maybe call this class `JsonDateSchema` because the time-part is optional? Your choice.

xcube/cli/_gen2/genconfig.py
```diff
@@ -207,8 +207,15 @@ def get_schema(cls):
                                             JsonNumberSchema(),
                                             JsonNumberSchema()]),
                 spatial_res=JsonNumberSchema(exclusive_minimum=0.0),
-                time_range=JsonArraySchema(items=[JsonStringSchema(format='date-time'),
-                                                  JsonStringSchema(format='date-time', nullable=True)]),
+                time_range=JsonArraySchema(items=[
```
forman, line 210: It looks as if we will need this code block in all our store impls and test code. You may create a utility class method in `JsonDatetimeSchema` class:

```python
@classmethod
def new_datetime_range(cls) -> JsonArraySchema: 
    return JsonArraySchema(items=[ ...
```

xcube/util/jsonschema.py
forman: Ok, then we'll leave it as-is.

forman (changes requested):
Good. Its only the CHANGES.md, I believe...

xcube/util/jsonschema.py
pont-us: I would rather keep it as Datetime, regarding a date as a special case of a date-time (even if that isn't true as regards the specification syntax). More correct would be JsonDateOrDatetimeSchema, but that's horribly unwieldy.

xcube/cli/_gen2/genconfig.py
pont-us: Done. (I used a static method rather than a class method, since it doesn't need access to the class.)

xcube/util/jsonschema.py
pont-us: I've added TODOs for anyOf handling in `from_instance` and `to_instance` and otherwise left it as is.

docs/source/opener_parameters_api.md
pont-us: Suggestion in previous comment now implemented.

PR #339, opened 2020-09-29
Introduce new dataset notation and cube type

TonioF:
Introduces new notation to denote types that have been derived from dataset and includes a new type ( dataset[cube] ) and a new DataDescriptor ( CubeDescriptor ).


forman (changes requested):
Besides my comments below, this implementation is incomplete. It does not reflect the semantics of the new `type_id` when passed to the following methods:

* [xcube.core.store.accessor.get_data_accessor_predicate()](https://github.com/dcs4cop/xcube/blob/master/xcube/core/store/accessor.py#L106)
* [xcube.core.store.store.DataStore.get_data_ids()](https://github.com/dcs4cop/xcube/blob/master/xcube/core/store/store.py#L104)
* [xcube.core.store.store.DataStore.search_data()](https://github.com/dcs4cop/xcube/blob/master/xcube/core/store/store.py#L104)
* [xcube.core.store.store.DataStore.get_data_opener_ids()](https://github.com/dcs4cop/xcube/blob/master/xcube/core/store/store.py#L104)
* [xcube.core.store.store.DataStore.get_data_writer_ids()](https://github.com/dcs4cop/xcube/blob/master/xcube/core/store/store.py#L261)

For the purpose of comparing two type_ids, I suggest we introduce a new class `TypeId`:

```python
class TypeId:

    def __init__(self, type_id: str):
        name, flags = self.parse_type_id(type_id)
        self._name: str = name
        self._flags: set = flags
 
    @property
    def name(self) -> str:
        return self._name

    @property
    def flags(self) -> Set[str]:
        return self._name

    def __eq__(self, other) -> bool:
        ...

    def __hash__(self) -> int:
        return hash(self.name) + 16 * hash(self.flags)

    @classmethod
    def normalize(cls, type_id: Union[str, "TypeId"]) -> TypeId:
        if instanceof(type_id, TypeId):
            return type_id
        if instanceof(type_id, str):
            return cls.parse(type_id)
        raise TypeError('type_id must be of type "str" or "TypeId"')

    @classmethod
    def parse(cls, type_id: str) -> TypeId:
         ..

```

And in the methods mentioned above, we allow `type_id` to be of either type `str` or `TypeId`: `Union[str, TypeId]`. To normalize, we call `TypeId.normalize(type_id)`.

pont-us (approved):
Approving since I have nothing to add to @forman 's review.

test/core/store/test_descriptor.py
```diff
@@ -8,30 +9,36 @@ class DatasetDescriptorTest(unittest.TestCase):
...
-        descriptor_dict = dict(data_id='xyz', type_id='abc')
+        descriptor_dict = dict(data_id='xyz')
         descriptor = DatasetDescriptor.from_dict(descriptor_dict)
         self.assertIsNotNone(descriptor)
         self.assertEqual('xyz', descriptor.data_id)
-        self.assertEqual('abc', descriptor.type_id)
+        self.assertEqual('dataset', descriptor.type_id)
+
+    def test_from_dict_derived_type(self):
+        descriptor_dict = dict(data_id='xyz', type_id='dataset[fegd]')
```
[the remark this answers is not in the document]
TonioF: Yes. This test only tests whether it is a dataset. It may hold any potential additional flags.

xcube/core/store/descriptor.py
```diff
@@ -219,6 +288,16 @@ def __init__(self,
         super().__init__(data_id=data_id, type_id=TYPE_ID_MULTI_LEVEL_DATASET, **kwargs)
         self.num_levels = num_levels
 
+    def _assert_type_id(self, type_id: str):
+        split_id = type_id.split('[')
+        if split_id[0] != 'dataset':
+            raise ValueError('Type ID must start with dataset')
```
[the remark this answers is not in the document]
TonioF: No. It must start with dataset and then hold the flag multilevel . (This might change when we agree to use the class TypeId as proposed above.)

xcube/core/store/descriptor.py
```diff
@@ -219,6 +288,16 @@ def __init__(self,
...
+        if len(split_id) == 1:
+            raise ValueError("Type ID must contain 'multilevel'")
```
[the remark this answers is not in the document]
TonioF: We are in the class MultiLevelDatasetDescriptor. This one definitely expects the flag multilevel to be set.

TonioF:
I have completely overhauled the pr. Now, the TypeId - class as proposed in @forman 's comment is used, and it is considered in the implementations of the stores.
MultilevelDataset is not considered a type of Dataset anymore, as they are two different classes and adhere to different interfaces. We now only have a flag cube to indicate that every cube is also a dataset.
I have further introduced a type id ANY_TYPE which corresponds to the asterisk which is used by some accessors to indicate that they work with any kind of data.

pont-us (approved):

test/core/store/test_typeid.py
```diff
@@ -0,0 +1,102 @@
...
+from xcube.core.store.typeid import TYPE_ID_DATASET
+from xcube.core.store.typeid import TYPE_ID_GEO_DATA_FRAME
+from xcube.core.store.typeid import TYPE_ID_MULTI_LEVEL_DATASET
+from xcube.core.store.typeid import get_type_id
+from xcube.core.store.typeid import TypeId
+
+from xcube.core.mldataset import BaseMultiLevelDataset
+
+
+class TypeIdTest(unittest.TestCase):
```
forman, line 16: In `TypeIdTest` test methods, please use names and flags that apply to our context, so we (and others) can better grasp the intention of `TypeId`.

xcube/core/store/descriptor.py
```diff
@@ -139,17 +130,20 @@ def __init__(self,
         self.data_vars = list(data_vars) if data_vars else None
         self.attrs = dict(attrs) if attrs else None
 
+    def _assert_type_id(self, type_id: str):
+        if not TYPE_ID_DATASET.is_compatible(type_id):
+            raise ValueError('TypeId must be "dataset" type id')
```
forman, line 135:
```suggestion
            raise ValueError('type_id must be compatible with "dataset"')
```

Note that value errors are used to signal developer errors, hence it should include the name of the actual variable that includes the invalid value (`type_id` here).

xcube/core/store/descriptor.py
```diff
@@ -219,6 +213,10 @@ def __init__(self,
         super().__init__(data_id=data_id, type_id=TYPE_ID_MULTI_LEVEL_DATASET, **kwargs)
         self.num_levels = num_levels
 
+    def _assert_type_id(self, type_id: str):
+        if not TYPE_ID_MULTI_LEVEL_DATASET.is_compatible(type_id):
+            raise ValueError('TypeId must be "mldataset" type id')
```
forman, line 218: Same as above.

xcube/core/store/descriptor.py
```diff
@@ -248,6 +246,10 @@ def __init__(self,
                          **kwargs)
         self.feature_schema = feature_schema
 
+    def _assert_type_id(self, type_id: str):
+        if not TYPE_ID_GEO_DATA_FRAME.is_compatible(type_id):
+            raise ValueError('TypeId must be "geodataframe" type id')
```
forman, line 251: Same as above.

xcube/core/store/typeid.py
```diff
@@ -0,0 +1,124 @@
...
+        if not type_id.endswith(']'):
+            raise SyntaxError(f'"{type_id}" cannot be parsed: No end brackets found')
+        name = type_id.split('[')[0]
+        flags = type_id.split('[')[1].split(']')[0].split(',')
+        return TypeId(name, flags=set(flags))
+
+TYPE_ID_ANY = TypeId('*')
+TYPE_ID_DATASET = TypeId('dataset')
+TYPE_ID_CUBE = TypeId('dataset', flags={'cube'})
+TYPE_ID_MULTI_LEVEL_DATASET = TypeId('mldataset')
```
forman, line 112: Actually, this is the correct definition:

```suggestion
TYPE_ID_MULTI_LEVEL_DATASET = TypeId('dataset', flags={'multilevel'})
```

However, a `MultiLevelDataset` is not yet an instance  of `xarray.Dataset` but that should be possible later.

xcube/core/store/typeid.py
```diff
@@ -0,0 +1,124 @@
...
+TYPE_ID_GEO_DATA_FRAME = TypeId('geodataframe')
+
+def get_type_id(data: Any) -> Optional[TypeId]:
+    if isinstance(data, xr.Dataset):
+        if 'time' in data.coords and 'lat' in data.coords and 'lon' in data.coords:
+            return TYPE_ID_CUBE
+        return TYPE_ID_DATASET
+    elif isinstance(data, MultiLevelDataset):
```
forman, line 120: See comment above.

xcube/core/store/typeid.py
```diff
@@ -0,0 +1,124 @@
...
+import xarray as xr
+
+class TypeId:
+    """
+    A type id denotes a type of data. It is used to group similar types of data and discern different types of data.
+    It can be used by stores to state what types of data can be read from and/or written to them.
+
+    A type id consists of a name and an arbitrary number of flags.
+    Flags can be used to further refine a type id if needed.
+    """
```
forman, line 40: Missing description of paramerters to `__init__()`.

forman (changes requested):
Very good, thanks for this!

We still need to discuss about how to deal with `MultiLevelDataset` which actually _is a_ `Dataset`, but not yet in Python terms. See my comments in `TypeId`.

xcube/core/store/descriptor.py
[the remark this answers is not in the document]
forman: See my new comments. MLDataset _is a_ dataset.

xcube/core/store/descriptor.py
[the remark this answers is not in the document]
forman: Still error message are not correctly formatted, see my new comments.

TonioF:
@forman , I have applied most of your change requests.
I do not feel too well with treating multileveldatasets as a sub-type of datasets already. E.g., if I have a DataOpener for Datasets, I would expect that I can open all types of datasets with them. I would use a MultilevelDataOpener if I want to ensure that what I open meets the criteria of a multileveldataset. This expectation would not be met if we already make MultilevelDatasets sub-types of datasets.
Having stated my concerns, do you still want me to make this change to the pr?

pont-us (approved):

PR #345, opened 2020-10-13
updated store notebooks and added notebook for cds

AliceBalfanz:
To review this PR, please take a look at:
examples/store/open_data_cciodp.ipynb 
examples/store/open_data_cds.ipynb 
examples/store/open_data_sentinelhub.ipynb


TonioF (approved):
Nice work in general, though I have a few recommendations for improvement. I put them here because (a) they mostly apply to all 3 notebooks and (b) commenting in the notebooks is a hassle.
- "This notebook walks provides a walk-through demonstrating how to use xcube and the xcube ... plugin to read and explore data from ... ." -> "This notebook provides a walk-through demonstrating how to use xcube and the xcube ... plugin to read and explore data from the ... ." in all 3 notebooks.
- For the CDS and SH notebooks: Consider inserting this cell which is a nifty way to set environment variables in Python:
import os
os.environ['MY_FIRST_ENVIRONMENT_VARIABLE'] = ''
- For the CDS Notebook: Mention that registering and using is free
- For the SH Notebook: The SH_INSTANCE_ID is also required
- For the CCI Notebook. Norman has implemented a nicer way to enable asyncio:
from xcube.util.ipython import enable_asyncio
enable_asyncio()
- "Usually we need more information to get the actual data store object. Which data store parameters are available?" -> "The data stores may offer different initialization options. Some of the options are mandatory. Let's have a look at the parameters for this store."
- "Which in-memory data types are provided?" -> I feel this is rather confusing right now as the three stores only offer datasets / All 3 notebooks
- "The first opener is the default opener. In this case there is only a single one:" Please also add this sentence to the CCI notebook
- In the SH notebook, you show a map. It would be nice to have some text with that. In general, there is very little text at the end of that notebook.

PR #346, opened 2020-10-21
Toniof update store conv

TonioF:
State that some opener parameters are mandatory.


docs/source/storeconv.md
```diff
@@ -76,7 +76,13 @@ be represented fully by splitting datasets in this way. In these cases:
 
 While an opener is free to define any open parameters for any of its datasets,
 there are some common parameters which are likely to be used by the majority
-of datasets. Any dataset requiring the specification of these parameters MUST
+of datasets. Also, there are some parameters which are fundamental for the 
+description of a dataset and therefore MUST be included in a schema 
+(these parameters are denoted explicitly in the list below). In case that an 
+opener does not support varying values of one of these parameters, a constant 
+value must be set.
```
forman, line 83:
```suggestion
opener does not support varying values of one of these parameters, a constant 
value must defined. This may be achieved by the JSON schema's `const` property 
or by an `enum` property whose value is a one-element array.
```

forman (changes requested):

PR #349, opened 2020-10-28
All stores to use JsonDateSchema.new_range()

forman:
In this PR:

* Introduced new JsonDateSchema and JsonComplexSchema.
* All data stores now use `time_range=JsonDateSchema.new_range()`, therefore no longer allowing date-time instances. This has been done to provide the xcube generator UI + service with a consistent cube configuration parameter.

### DIFF
--- a/xcube/util/jsonschema.py
+++ b/xcube/util/jsonschema.py
@@ -22,6 +22,9 @@
 from abc import ABC, abstractmethod
 from typing import Dict, Any, Callable, Mapping, Sequence, Union, Tuple, Optional
 
+# Make sure strict-rfc3339 package is installed. The package jsonschema uses it for validating
+# instances of the JsonDateSchema and JsonDatetimeSchema.
+__import__('strict_rfc3339')
 import jsonschema
 
 from xcube.util.ipython import register_json_formatter
@@ -216,7 +219,7 @@ class JsonDateAndTimeSchemaBase:
 
     # noinspection PyShadowingBuiltins
     @classmethod
-    def _is_valid_value(cls, value: str, format: str):
+    def _is_valid_value(cls, value: str, format: str) -> bool:
         try:
             jsonschema.validate(value,
                                 dict(type='string', format=format),
@@ -287,8 +290,8 @@ class JsonDatetimeSchema(JsonStringSchema, JsonDateAndTimeSchemaBase):
                  max_datetime: str = None,
                  **kwargs):
         super().__init__(**kwargs, format='date-time')
-        JsonDateSchema._validate_value(min_datetime, 'min_datetime', 'date-time')
-        JsonDateSchema._validate_value(max_datetime, 'max_datetime', 'date-time')
+        self._validate_value(min_datetime, 'min_datetime', 'date-time')
+        self._validate_value(max_datetime, 'max_datetime', 'date-time')
         self.min_datetime = min_datetime
         self.max_datetime = max_datetime
 
--- a/xcube/util/jsonschema.py
+++ b/xcube/util/jsonschema.py
@@ -20,7 +20,7 @@
 # SOFTWARE.
 
 from abc import ABC, abstractmethod
-from typing import Dict, Any, Callable, Mapping, Sequence, Union, Tuple
+from typing import Dict, Any, Callable, Mapping, Sequence, Union, Tuple, Optional
 
 import jsonschema
 
@@ -36,14 +36,9 @@ _NUMERIC_TYPES_ENUM = {'integer', 'number'}
 
 class JsonSchema(ABC):
 
-    # TODO: implement any_of more completely
-    # Currently any_of is only used by JsonDatetimeSchema to allow alternative
-    # 'date' and 'date-time' format specifiers. For full support it should also
-    # be handled in from_instance and to_instance.
-
     # noinspection PyShadowingBuiltins
     def __init__(self,
-                 type: str,
+                 type: Union[str, Sequence[str]] = None,
                  default: Any = UNDEFINED,
                  const: Any = UNDEFINED,
                  enum: Sequence[Any] = None,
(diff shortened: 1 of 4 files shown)


xcube/util/jsonschema.py
```diff
@@ -175,54 +206,91 @@ def to_dict(self) -> Dict[str, Any]:
...
+    @classmethod
+    def new_range(cls,
+                  min_date: str = None,
+                  max_date: str = None,
+                  nullable: bool = False) -> 'JsonArraySchema':
+        """
+        Return a schema for a date range.
+        :param min_date: optional minimum date.
+        :param max_date: optional maximum date.
+        :param nullable: whether the whole range as well as individual start and end may be None.
```
TonioF, line 265: -> "individual start and end DATES"

xcube/util/jsonschema.py
```diff
@@ -175,54 +206,91 @@ def to_dict(self) -> Dict[str, Any]:
...
+    :param max_datetime: optional maximum date-time.
+    """
+
+    # noinspection PyShadowingBuiltins
+    def __init__(self,
+                 min_datetime: str = None,
+                 max_datetime: str = None,
+                 **kwargs):
+        super().__init__(**kwargs, format='date-time')
+        JsonDateSchema._validate_value(min_datetime, 'min_datetime', 'date-time')
```
TonioF, line 290: In Json DateSchema, you specify this as self._validate_value(...) . The same function is used, but I'd prefer to see it called the same way.

TonioF (changes requested):
"test_min_max_validity_checks" from JsonDatetimeSchemaTest fails.

I would expect JsonDateSchema and JsonDatetimeSchema to describe single date instances, even their docs indicate that. Instead, they both describe ranges. I suggest to either add the word "range" to their name (and change the docs) or to introduce different classes for single dates and ranges. I think the former is the better solution.

Why was it necessary to create different classes for JsonDateSchema and JsonDatetimeSchema? Why wasn't it possible to just have two different functions new_date_range and new_date_time_range or just have a single method with a mandatory format parameter? JsonDateSchema and JsonDatetimeSchema now have almost the same code.

pont-us:
> "test_min_max_validity_checks" from JsonDatetimeSchemaTest fails.

It only fails if the `strict-rfc3339` package is not present. In this case, the `jsconschema` package can't check the date format, so passes the strings as valid (which is the officially mandated JSON Schema behaviour in this case). One possible solution would be to skip the test if `strict-rfc3339` is missing. Another would be for `xcube.jsonschema` to enforce a dependency on `strict-rfc3339`, perhaps simply by trying to import somethine from it. (Personally I think the latter solution would be preferable.)

pont-us:
> Instead, they both describe ranges.

No, a `JsonDateSchema` or `JsonDatetimeSchema` describes a single value (optionally) restricted to a specified range, analogous to e.g. `{"type": "number", "minimum": 1, "maximum": 10}`.

xcube/util/jsonschema.py
pont-us: It will in any case call through to `JsonDateAndTimeSchemaBase`, but I agree that `JsonDatetimeSchema` going via `JsonDateSchema` to call `_validate_value` is rather awkward. Either `self._validate_value` or `JsonDatetimeSchema._validate_value` would be preferable (personally I prefer the latter since it makes it clear that a class method is being called).

pont-us (changes requested):
I'd like to have the [soft strict-rfc3339 dependency](https://github.com/dcs4cop/xcube/pull/349#issuecomment-718733020) dealt with somehow before merging.

xcube/util/jsonschema.py
```diff
@@ -175,54 +206,91 @@ def to_dict(self) -> Dict[str, Any]:
...
+        """
+        Return a schema for a date range.
+        :param min_date: optional minimum date.
+        :param max_date: optional maximum date.
+        :param nullable: whether the whole range as well as individual start and end may be None.
+        :return: a JsonArraySchema with two items.
+        """
+        return JsonArraySchema(
+            items=[
+                JsonDateSchema(min_date=min_date, max_date=max_date, nullable=nullable),
```
pont-us, line 270:
```suggestion
                JsonDateSchema(min_date=min_date, max_date=max_date, nullable=True),
```

The documentation says that `min_date` is optional, so its schema should be nullable whether or not the enclosing array schema is nullable.

xcube/util/jsonschema.py
```diff
@@ -175,54 +206,91 @@ def to_dict(self) -> Dict[str, Any]:
...
+                JsonDateSchema(min_date=min_date, max_date=max_date, nullable=nullable),
```
pont-us, line 271:
```suggestion
                JsonDateSchema(min_date=min_date, max_date=max_date, nullable=True),
```

The documentation says that `max_date` is optional, so its schema should be nullable whether or not the enclosing array schema is nullable.

pont-us (changes requested):

xcube/util/jsonschema.py
forman: Happened by accident.

xcube/util/jsonschema.py
forman: > I would expect JsonDateSchema and JsonDatetimeSchema to describe single date instances, even their docs indicate that. Instead, they both describe ranges.

?

forman: min/max are used to validate instances. Please familiarize yourself with JSON schema first.

xcube/util/jsonschema.py
forman: > The documentation says that min_date is optional, so its schema should be nullable whether or not the enclosing array schema is nullable.

Why do you think nullable depends on min_date ?

xcube/util/jsonschema.py
forman: Same as above

forman:
Merging after PR review call...

PR #351, opened 2020-10-30
reopening PR: updated store notebooks and added notebook for cds #345

AliceBalfanz:
To review this PR, please take a look at:
examples/store/open_data_cciodp.ipynb
examples/store/open_data_cds.ipynb
examples/store/open_data_sentinelhub.ipynb

Note: 
This also includes changes based on [comments from TonioF](https://github.com/dcs4cop/xcube/pull/345#pullrequestreview-510611312) - please view the original comment, because the copy of it here below is badly formatted :) : 

 

>   Nice work in general, though I have a few recommendations for improvement. I put them here because (a) they mostly apply to       all 3 notebooks and (b) commenting in the notebooks is a hassle.
> 
> 
>     "This notebook walks provides a walk-through demonstrating how to use xcube and the xcube ... plugin to read and explore data from ... ." -> "This notebook provides a walk-through demonstrating how to use xcube and the xcube ... plugin to read and explore data from the ... ." in all 3 notebooks.
>     For the CDS and SH notebooks: Consider inserting this cell which is a nifty way to set environment variables in Python:
>     import os
>     os.environ['MY_FIRST_ENVIRONMENT_VARIABLE'] = ''
>     For the CDS Notebook: Mention that registering and using is free
>     For the SH Notebook: The SH_INSTANCE_ID is also required
>     For the CCI Notebook. Norman has implemented a nicer way to enable asyncio:
>     from xcube.util.ipython import enable_asyncio
>     enable_asyncio()
>     "Usually we need more information to get the actual data store object. Which data store parameters are available?" -> "The data stores may offer different initialization options. Some of the options are mandatory. Let's have a look at the parameters for this store."
>     "Which in-memory data types are provided?" -> I feel this is rather confusing right now as the three stores only offer datasets / All 3 notebooks
>     "The first opener is the default opener. In this case there is only a single one:" Please also add this sentence to the CCI notebook
>     In the SH notebook, you show a map. It would be nice to have some text with that. In general, there is very little text at the end of that notebook.
>


pont-us (changes requested):
Suggested changes:

**Sentinel Hub notebook**

 - "SENTINEL" -> "Sentinel" throughout
 - envrionment -> environment in second paragraph
 - In first code cell, `import os` required before setting variables
 - In "Get more info about a dedicated dataset.", change "dedicated" -> "specific"
 - "This include a description": "include" -> "includes"

**CCI Notebook**

 - First paragraph: "Protal" -> "Portal"
 - I can't get the notebook to run correctly. With the latest version of the
   CCI store (master branch from the repo), calls to `search_data` and
   `open_data` result in:
   
```
~/loc/repos/xcube-cci/xcube_cci/dataaccess.py in _get_data_descriptor_from_metadata(self, data_id, ds_metadata)
    149             else:
    150                 var_descriptors.append(VariableDescriptor(var_name, '', ''))
--> 151         ds_metadata.pop('variables')
    152         ds_metadata.pop('dimensions')
    153         ds_metadata.pop('variable_infos')

KeyError: 'variables'
```

**CDS Notebook**

 - As in SH notebook, `import os` before the commented-out `os.environ`
   calls.

TonioF:
> * I can't get the notebook to run correctly. With the latest version of the
>   CCI store (master branch from the repo), calls to `search_data` and
>   `open_data` result in:
> 
> ```
> ~/loc/repos/xcube-cci/xcube_cci/dataaccess.py in _get_data_descriptor_from_metadata(self, data_id, ds_metadata)
>     149             else:
>     150                 var_descriptors.append(VariableDescriptor(var_name, '', ''))
> --> 151         ds_metadata.pop('variables')
>     152         ds_metadata.pop('dimensions')
>     153         ds_metadata.pop('variable_infos')
> 
> KeyError: 'variables'
> ```

This was an error in the xcube-cci store. It has been fixed in the time you made your review.

environment.yml
```diff
@@ -38,6 +38,7 @@ dependencies:
   - zarr >=2.4
   # Testing
   - flake8 >=3.7
+  - jupyter >=1.0
```
forman, line 41: Take this out. This is not an xcube requirement.

forman (changes requested):
Please fix:

- [x] `jupyter` must not be an xcube requirement! Please fix `environment.yml`. It adds a lot of transitive package dependencies to the xcube env.
- [ ] move `examples/store` into dedicated `examples/notebooks/store`

Let's fix this together in a dedicated TC:

- [ ] missing general NB that explains the top-level store functions (e.g. use directory store)
- [ ] rename notebooks as they refer to much more than "open"
- [ ] missing explanation where store plugins come from 
- [ ] explain required imports and import used only for demonstration purpose
- [ ] "data accessor" --> "data store" in NB text cells
- [ ] "initialization options" vs. "parameters for this store" in text cells
- [ ] "Which "openers" are available?" the explanation is outdated and fix should go into general NB

forman:
> * "SENTINEL" -> "Sentinel" throughout

In the case of "SENTINEL Hub" this a proper name written exactly like so. At least at that time, they may have changed it.

forman:
> > * "SENTINEL" -> "Sentinel" throughout

> At least at that time, they may have changed it. 

They did. I think we are save with "Sentinel Hub". See https://www.sentinel-hub.com/explore/

pont-us:
> They did. I think we are save with "Sentinel Hub". See https://www.sentinel-hub.com/explore/

Yes, I checked their own current usage before making the request :).

forman (approved):
Thanks, Alicja!.

PR #353, opened 2020-11-02
S3 tests fix

AliceBalfanz:
The PR addresses issues with s3 tests on travis - on one hand the dataset used for testing is now a spatial, variable and time subset of the initial dataset, it also changes slightly the use of the s3-test setup of moto server. This PR contains many contributions by forman. 
Implementation within xcubes code itself has not been changed.


.travis.yml
```diff
@@ -5,6 +5,9 @@
 
 language: generic
 
+#env:
```
forman, line 8:
```suggestion
# Turn PARALLELIZE=true only after the build passed with PARALLELIZE=false.
# Turn PARALLELIZE=false as a first action after tests fail, especially those that use 
# concurrent processes or threads (e.g. S3 moto server)!
#
#env:
```

forman (changes requested):
Please use my suggestion that explains use of `PARALLELIZE`.

Issue #355, opened 2020-11-03
Readthedocs compilation not complete


AliceBalfanz:
**Describe the bug**
The documentation on Readthedocs is not listing all configured docstrings. The issue does not exist when compiling the documentation locally within the `xcube/docs` directory with `make html`. 

read the docs of master branch: https://xcube.readthedocs.io/en/latest/api.html 
rst file with autofunctions pointing to the modules: https://github.com/dcs4cop/xcube/blob/master/docs/source/api.rst

When checking the Read the docs build for the latest code version on https://readthedocs.org/ the step:

` python /home/docs/checkouts/readthedocs.org/user_builds/xcube/envs/latest/bin/sphinx-build -T -E -b readthedocs -d _build/doctrees -D language=en . _build/html`

produces many warnings. The xcube modules with warnings do not show the docstrings within the read the docs documentation: 
```
Running Sphinx v3.2.1
loading translations [en]... done
making output directory... done
building [mo]: targets for 0 po files that are out of date
building [readthedocs]: targets for 30 source files that are out of date
updating environment: [new config] 30 added, 0 changed, 0 removed
reading sources... [  3%] api
reading sources... [  6%] cli
reading sources... [ 10%] cli/xcube
reading sources... [ 13%] cli/xcube_chunk
reading sources... [ 16%] cli/xcube_compute
reading sources... [ 20%] cli/xcube_dump
reading sources... [ 23%] cli/xcube_edit
reading sources... [ 26%] cli/xcube_extract
reading sources... [ 30%] cli/xcube_gen
reading sources... [ 33%] cli/xcube_grid
reading sources... [ 36%] cli/xcube_level
reading sources... [ 40%] cli/xcube_optimize
reading sources... [ 43%] cli/xcube_prune
reading sources... [ 46%] cli/xcube_resample
reading sources... [ 50%] cli/xcube_serve
reading sources... [ 53%] cli/xcube_tile
reading sources... [ 56%] cli/xcube_vars2dim
reading sources... [ 60%] cli/xcube_verify
reading sources... [ 63%] cubespec
reading sources... [ 66%] devguide
reading sources... [ 70%] examples
reading sources... [ 73%] examples/xcube_gen
reading sources... [ 76%] examples/xcube_serve
reading sources... [ 80%] index
reading sources... [ 83%] installation
reading sources... [ 86%] overview
reading sources... [ 90%] plugins
reading sources... [ 93%] storeconv
reading sources... [ 96%] viewer
[...]

PR #356, opened 2020-11-05
Included missing packages to mock imports in conf.py for rtd 

AliceBalfanz:
Read the docs was not compiling all doc strings due to missing imports within the autodoc_mock_imports of conf.py
I have added the missing modules and sorted the existing list alphabetically. Within config the extensions have been modified as well as pandas and jsonschema added to the requirements in order to fix all errors which appeared within the rtd build. 

Check out the docu: https://xcube.readthedocs.io/en/alicja-xxx-fix-docs/api.html#cube-computation

this is how it was before: 

image

this addresses issue https://github.com/dcs4cop/xcube/issues/355


forman (approved):
Supi! Vielen Dank!

PR #358, opened 2020-11-06
changes of storeconv.md

AliceBalfanz:
While reading the Store conventions, I stumbled across some sentences and suggested changes in this PR.

### DIFF
--- a/docs/source/storeconv.md
+++ b/docs/source/storeconv.md
@@ -22,7 +22,7 @@ Useful references related to this document include:
 
 ## Naming Identifiers
 
-This section states how identifiers of an xcube data store SHALL look like.
+This section defines the structure of the identifiers used by an xcube data store.
 
 Within the DataStore framework, identifiers are used to denote data sources, data stores, and data accessors.
 Data store, data opener, and data writer identifiers are used to register the component as an extension in a 
@@ -172,7 +172,7 @@ annotations.
 
 ### Semantics of list-valued parameters
 
-The `variables` parameter takes as its value a list, with no members duplication
+The `variables` parameter takes as its value a list, with no duplicated members
 and the values of its members drawn from a predefined set. The values of this
 parameter, and other parameters which values also follow such a format, are
 interpreted by xcube as a *restriction*, much like a bounding box or time
--- a/docs/source/storeconv.md
+++ b/docs/source/storeconv.md
@@ -55,7 +55,7 @@ Examples for valid data accessors identifiers are:
 
 Type Specifiers are used to specify a data type.
 They consist of a name and an arbitrary number of optional flags, given in square brackets. 
-These flags are used to define characteristics of a type, e.g., the type specifier `dataset[cube]` denotes a dataset 
+These flags are used to define characteristics of a type, e.g., the type specifier "`dataset[cube]`" denotes a dataset 
 which also meets the requirements of a cube. 
 A dataset specified by `dataset[cube, multilevel]` is a cube and has multiple levels.
 The order of flags is irrelevant, i.e., `dataset[cube, multilevel]` is the same as `dataset[multilevel, cube]`. 
--- a/docs/source/storeconv.md
+++ b/docs/source/storeconv.md
@@ -24,9 +24,8 @@ Useful references related to this document include:
 
 This section explains various identifiers used by the xcube data store framework and defines their format.
 
-Within the DataStore framework, identifiers are used to denote data sources, data stores, and data accessors.
-Data store, data opener, and data writer identifiers are used to register the component as an extension in a 
-package's plugin.py.
+In the DataStore framework, identifiers are used to denote data sources, data stores, and data accessors.
+Data store, data opener, and data writer identifiers are used to register the component as an extension in a package's plugin.py.
 Identifiers MUST be unambiguous within the scope of the data store. 
 They SHOULD be unambiguous across the entirety of data stores. 
 
@@ -97,11 +96,11 @@ interface for accessing the available datasets. Note that, under this system:
 To work around some of the restrictions of point (2) above, a dataset MAY be
 presented by the opener as multiple "virtual" datasets with different
 parameter schemas. For instance, the hypothetical dataset described above MAY 
-be offered as two datasets
+be offered not as a single dataset `envdata` but as two datasets
 `envdata:sensor-a` (with a fixed *sensor_type* of A) and `envdata:sensor-b`,
 (with a fixed *sensor_type* of B), offering different sets of permitted
-variables.
-
+ variables. 
+ 
 Sometimes, the interdependencies between parameters are too complex to
 be fully represented by splitting datasets in this manner. In these cases:
 
@@ -124,7 +123,7 @@ description of a dataset and therefore MUST be included in a schema
 (these parameters are denoted explicitly in the list below). In case that an 
 opener does not support varying values of one of these parameters, a constant 
 value must defined. This may be achieved by the JSON schema's `const` property 
-or by an `enum` property value which is a one-element array.
+or by an `enum` property value whose is a one-element array.
 
 Any dataset requiring the specification of these parameters MUST
 use the standard parameter names, syntax, and semantics defined below, in
@@ -174,7 +173,7 @@ annotations.
 
 The `variables` parameter takes as its value a list, with no duplicated members
 and the values of its members drawn from a predefined set. The values of this
-parameter, and other parameters which values also follow such a format, are
+parameter, and other parameters whose values also follow such a format, are
 interpreted by xcube as a *restriction*, much like a bounding box or time
 range. That is:
 
@@ -192,7 +191,7 @@ range. That is:
 
 ### <a id="sec-datespec"></a>Date, time, and duration specifications
 
-Within the common parameter `time_range`, times can be specified using the
+In the common parameter `time_range`, times can be specified using the
 standard JSON Schema formats `date-time` or `date`. Any additional time or
 date parameters supported by an xcube opener dataset SHOULD also use these
 formats, unless there is some good reason to prefer a different format.
--- a/docs/source/storeconv.md
+++ b/docs/source/storeconv.md
@@ -22,11 +22,12 @@ Useful references related to this document include:
 
 ## Naming Identifiers
 
-This section states what identifiers used by an xcube data store SHALL look like.
+This section states how identifiers of an xcube data store SHALL look like.
 
-In the DataStore framework, identifiers are used to denote data sources, data stores, and data accessors.
-Data store, data opener, and data writer identifiers are used to register the component as extension in a package's plugin.py.
-Identifiers MUST be unambiguous in the scope of the data store. 
+Within the DataStore framework, identifiers are used to denote data sources, data stores, and data accessors.
+Data store, data opener, and data writer identifiers are used to register the component as an extension in a 
+package's plugin.py.
+Identifiers MUST be unambiguous within the scope of the data store. 
 They SHOULD be unambiguous across the entirety of data stores. 
 
 There are no further restrictions for data source and data store identifiers.
@@ -37,7 +38,8 @@ A data accessor identifier MUST correspond to the following scheme:
 
 `<type_specifier>` specifies a data type. 
 Its intention and format is described in the sub-section below.
-In case the type specifier has flags, the flags MUST be given in square brackets, in alphabetic order, separated by single commas, without spaces (e.g., `dataset[cube,multilevel]`).
+In case the type specifier has flags, the flags MUST be given within square brackets, in alphabetic order, separated by 
+single commas and without spaces (e.g., `dataset[cube,multilevel]`).
 The `<format_identifier>` describes the data format that may be accessed, e.g., `zarr` or `netcdf`.
 The `<storage_identifier>` describes the kind of storage or data provision the accessor can access.
 Example values are `memory`, `s3` or `sentinelhub`.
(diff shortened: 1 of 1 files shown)


docs/source/storeconv.md
```diff
@@ -22,11 +22,12 @@ Useful references related to this document include:
 
 ## Naming Identifiers
 
-This section states what identifiers used by an xcube data store SHALL look like.
+This section states how identifiers of an xcube data store SHALL look like.
```
pont-us, line 25: I don't find"of" appropriate here; it misleadingly implies that these identifiers only apply to stores.

"how" is ungrammatical. "How does it look?" and "What does it look like?" are correct sentences, but not "How does it look like?". But in any case I was never entirely happy with the slightly vague use of "look" here. I would suggest something like: "This section defines the structure of the identifiers used by an xcube data store".

docs/source/storeconv.md
```diff
@@ -74,11 +78,11 @@ Every implementation of the `xcube.core.store.DataOpener` or
 allowed arguments to `open_data` for each dataset supported by the
 `DataOpener` or `DataStore`. The description is provided as a
 `JsonObjectSchema` object corresponding to a [JSON
-Schema](https://json-schema.org/). The intention is that this description be
-full and detailed enough to allow the automatic construction of a user
-interface for access to the available datasets. Note that, under this system:
+Schema](https://json-schema.org/). The intention is that this description should be 
+detailed enough to allow the automatic construction of a user
+interface for accessing the available datasets. Note that, under this system:
```
pont-us, line 83: I'm not convinced that the rephrasing is an improvement... but perhaps after all it's safest to avoid the subjunctive  in technical documents, so I won't object :).

docs/source/storeconv.md
```diff
@@ -92,14 +96,14 @@ interface for access to the available datasets. Note that, under this system:
...
 To work around some of the restrictions of point (2) above, a dataset MAY be
 presented by the opener as multiple "virtual" datasets with different
-parameter schemas. For instance, the hypothetical dataset described above MAY
-be offered not as a single dataset `envdata` but as two datasets
+parameter schemas. For instance, the hypothetical dataset described above MAY 
+be offered as two datasets
 `envdata:sensor-a` (with a fixed *sensor_type* of A) and `envdata:sensor-b`,
 (with a fixed *sensor_type* of B), offering different sets of permitted
-variables.
+variables. However, it may not be represented as a single dataset `envdata`.
```
pont-us, line 103:
```suggestion
variables.
```

No, that's not what I meant! It's also allowed to offer a single dataset. Probably best to remove this clause entirely to avoid misunderstanding.

docs/source/storeconv.md
```diff
@@ -115,56 +119,62 @@ be represented fully by splitting datasets in this way. In these cases:
...
 While an opener is free to define any open parameters for any of its datasets,
 there are some common parameters which are likely to be used by the majority
-of datasets. Also, there are some parameters which are fundamental for the 
+of datasets. Furthermore, there are some parameters which are fundamental for the 
 description of a dataset and therefore MUST be included in a schema 
 (these parameters are denoted explicitly in the list below). In case that an 
 opener does not support varying values of one of these parameters, a constant 
 value must defined. This may be achieved by the JSON schema's `const` property 
-or by an `enum` property whose value is a one-element array.
+or by an `enum` property value which is a one-element array.
```
pont-us, line 127: I find this less clear than the original. If you don't like "whose", I suggest:

> ... or by an `enum` property with a one-element array as its value.

docs/source/storeconv.md
```diff
@@ -115,56 +119,62 @@ be represented fully by splitting datasets in this way. In these cases:
...
    parameter set to `True`, the store MUST return a specification-compliant
-   xcube. If this parameter is not supported or if a dataset is opened with
+   xcube dataset. If this parameter is not supported or if a dataset is opened with
    this parameter set to `False`, the caller MUST NOT assume that the returned
    data conform to the xcube specification.
 
 ### Semantics of list-valued parameters
 
-The `variables` parameter takes as its value a list, with no members repeated
+The `variables` parameter takes as its value a list, with no members duplication
```
pont-us, line 175: Ungrammatical. "member duplication" (without "s") would be grammatically correct, but I would find "duplicated members" clearer.

docs/source/storeconv.md
```diff
@@ -115,56 +119,62 @@ be represented fully by splitting datasets in this way. In these cases:
...
 and the values of its members drawn from a predefined set. The values of this
-parameter, and other parameters whose values also follow such a format, are
+parameter, and other parameters which values also follow such a format, are
```
pont-us, line 177: Ungrammatical. The alternative to "whose" would be "... other parameters, the values of which also follow such a format, ...", but I find that rather clunky.

docs/source/storeconv.md
```diff
@@ -176,20 +186,20 @@ range. That is:
    corresponding to those members only MUST be returned. In the case of
    `variables`, this will result in a dataset containing only the requested
    variables.
- - As a special case of the above, if an empty list is supplied, a dataset
+ - A special case of the above: If an empty list is supplied, a dataset
```
pont-us, line 189:
```suggestion
 - A special case of the above: if an empty list is supplied, a dataset
```

docs/source/storeconv.md
```diff
@@ -176,20 +186,20 @@ range. That is:
...
    containing *no data* MUST be returned -- but with the requested spatial and
    temporal dimensions.
 
 ### <a id="sec-datespec"></a>Date, time, and duration specifications
 
-In the common parameter `time_range`, times can be specified using the
+Within the common parameter `time_range`, times can be specified using the
```
pont-us, line 195: I'm not sure that this makes things clearer. On consideration, my "can" is also rather vague for a specification document, giving the impression that it's optional. I suggest rewriting this sentence as:

 - In values for the parameter `time_range`, the string components MUST conform to either the `date-time` or `time` JSON schema format.

pont-us (changes requested):

docs/source/storeconv.md
AliceBalfanz: 'Who is always associated with people. Which is used with things. ' https://www.grammarly.com/blog/when-to-use-which-and-who/
That's why I object to using 'whose' in this situation.

docs/source/storeconv.md
AliceBalfanz: Disagree for the usage of whose, because:  'Who is always associated with people. Which is used with things. ' https://www.grammarly.com/blog/when-to-use-which-and-who/

docs/source/storeconv.md
forman: I agree with Pont.

docs/source/storeconv.md
```diff
@@ -22,11 +22,12 @@ Useful references related to this document include:
 
 ## Naming Identifiers
 
-This section states what identifiers used by an xcube data store SHALL look like.
+This section defines the structure of the identifiers used by an xcube data store.
```
forman, line 25:
```suggestion
This section explains various identifiers used by the xcube data store framework and defines their format.
```

docs/source/storeconv.md
```diff
@@ -22,11 +22,12 @@ Useful references related to this document include:
...
 
-In the DataStore framework, identifiers are used to denote data sources, data stores, and data accessors.
-Data store, data opener, and data writer identifiers are used to register the component as extension in a package's plugin.py.
-Identifiers MUST be unambiguous in the scope of the data store. 
+Within the DataStore framework, identifiers are used to denote data sources, data stores, and data accessors.
+Data store, data opener, and data writer identifiers are used to register the component as an extension in a 
+package's plugin.py.
+Identifiers MUST be unambiguous within the scope of the data store.
```
forman, line 30: Not really an improvement. I believe original description if fine.

docs/source/storeconv.md
```diff
@@ -37,7 +38,8 @@ A data accessor identifier MUST correspond to the following scheme:
 
 `<type_specifier>` specifies a data type. 
 Its intention and format is described in the sub-section below.
-In case the type specifier has flags, the flags MUST be given in square brackets, in alphabetic order, separated by single commas, without spaces (e.g., `dataset[cube,multilevel]`).
+In case the type specifier has flags, the flags MUST be given within square brackets, in alphabetic order, separated by 
+single commas and without spaces (e.g., `dataset[cube,multilevel]`).
```
forman, line 42:
```suggestion
single commas and without spaces (e.g., `"dataset[cube,multilevel]"`).
```

docs/source/storeconv.md
```diff
@@ -53,10 +55,12 @@ Examples for valid data accessors identifiers are:
 
 Type Specifiers are used to specify a data type.
 They consist of a name and an arbitrary number of optional flags, given in square brackets. 
-These flags are used to define characteristics of a type, e.g., the type specifier `dataset[cube]` denotes a dataset which also meets the requirements of a cube. 
+These flags are used to define characteristics of a type, e.g., the type specifier `dataset[cube]` denotes a dataset
```
forman, line 58:
```suggestion
These flags are used to define characteristics of a type, e.g., the type specifier "`dataset[cube]`" denotes a dataset 
```

docs/source/storeconv.md
forman: I agree with Pont.

docs/source/storeconv.md
```diff
@@ -74,11 +78,11 @@ Every implementation of the `xcube.core.store.DataOpener` or
...
 `JsonObjectSchema` object corresponding to a [JSON
-Schema](https://json-schema.org/). The intention is that this description be
-full and detailed enough to allow the automatic construction of a user
-interface for access to the available datasets. Note that, under this system:
+Schema](https://json-schema.org/). The intention is that this description should be 
+detailed enough to allow the automatic construction of a user
+interface for accessing the available datasets. Note that, under this system:
 
-  1. Every dataset which an opener provides can support a different set of
+  1. Every dataset provided by an opener can support a different set of
```
forman, line 85: Better!

docs/source/storeconv.md
```diff
@@ -92,14 +96,14 @@ interface for access to the available datasets. Note that, under this system:
 
 To work around some of the restrictions of point (2) above, a dataset MAY be
 presented by the opener as multiple "virtual" datasets with different
-parameter schemas. For instance, the hypothetical dataset described above MAY
-be offered not as a single dataset `envdata` but as two datasets
+parameter schemas. For instance, the hypothetical dataset described above MAY 
+be offered as two datasets
```
forman, line 100: Less clear than the original. Should remain as it was.

docs/source/storeconv.md
```diff
@@ -92,14 +96,14 @@ interface for access to the available datasets. Note that, under this system:
...
 `envdata:sensor-a` (with a fixed *sensor_type* of A) and `envdata:sensor-b`,
 (with a fixed *sensor_type* of B), offering different sets of permitted
 variables.
 
 Sometimes, the interdependencies between parameters are too complex to
-be represented fully by splitting datasets in this way. In these cases:
+be fully represented by splitting datasets in this manner. In these cases:
```
forman, line 106: Ok.

docs/source/storeconv.md
forman: That is not correct: "whose" does not necessarily refer to a "who" as a person, it may refer to this things as well (de: deren / dessen). See https://www.linguee.de/deutsch-englisch/search?source=auto&query=whose

Therefore please stay with the original.

docs/source/storeconv.md
```diff
@@ -115,56 +119,62 @@ be represented fully by splitting datasets in this way. In these cases:
...
-or by an `enum` property whose value is a one-element array.
+or by an `enum` property value which is a one-element array.
 
 Any dataset requiring the specification of these parameters MUST
 use the standard parameter names, syntax, and semantics defined below, in
 order to keep the interface consistent. For instance, if a dataset allows a
 time aggregation period to be specified, it MUST use the `time_period`
 parameter with the format described below rather than some other alternative
-name and/or format. The parameters are described below with their Python type
+name and/or format. Below, the parameters are described with their Python type
```
forman, line 134: Ok.

docs/source/storeconv.md
forman: Same as abobe. Pont is right, please leave as-is.

docs/source/storeconv.md
forman: Stay with original.

forman (changes requested):
A number of change suggestions should not be merged as they are not really an improvement over the original.

docs/source/storeconv.md
TonioF: Maybe lose the "the" in Pont's sentence, otherwise I prefer his sentence.

docs/source/storeconv.md
TonioF: Please maintain word order: "... an enum property WHOSE value ..."

docs/source/storeconv.md
TonioF: Or just "duplicate members"

TonioF (changes requested):
Added my review as comments. In one case, there is a change necessary, the others are suggestions.

docs/source/storeconv.md
pont-us: "Who" is not the same as "whose". [Wikipedia has a whole article](https://en.wikipedia.org/wiki/Inanimate_whose) on "inanimate whose", and doesn't list a single modern style guide that recommends against it.

> *Merriam–Webster's Dictionary of English Usage* states that, amongst "the current books" that discussed the subject as of the late 20th century, "not one of them finds [inanimate] whose anything but standard". … It asserts that "notion that whose may not properly be used of anything except persons is a superstition" and such use is "entirely standard as an alternative to of which in all varieties of discourse".

[grammarly.com seems to feel the same way](https://www.grammarly.com/blog/whos-whose/):

> > Bessie carried a lantern, whose light glanced on wet steps and gravel road sodden by a recent thaw.

> This one’s worth an extra glance: “who” in all its forms generally refers to animate beings, but in the possessive there’s no equivalent for inanimate objects like Bessie’s lantern. The very awkward alternative is “Bessie carried a lantern, the light of which glanced on wet steps.” Not good.

docs/source/storeconv.md
pont-us: See my answer in previous discussion :).

PR #366, opened 2020-11-13
Update and expand installation guide

### LINKED ISSUE #363: Update installation guide and suggest using mamba
The [installation section](https://xcube.readthedocs.io/en/latest/installation.html) of the xcube user guide recommends installation using the conda command-line tool. In practice, this can be extremely slow (several hours). We should update the documentation to suggest or recommend the use of [mamba](https://github.com/mamba-org/mamba), a dramatically faster drop-in replacement for conda.

pont-us:
Fixes #363 .

 - Add more explanatory text.

 - Add a section describing use of mamba.

 - Fix a couple of buggy commands.

### DIFF
--- a/docs/source/installation.md
+++ b/docs/source/installation.md
@@ -1,50 +1,94 @@
-
 # Installation
 
-## Installation using conda
+xcube can be installed from a released conda package, or directly from a
+copy of the source code repository.
 
-Into existing conda environment (>= Python 3.7)
+The first two sections below give instructions for installation using conda,
+available as part of the [miniconda
+distribution](https://docs.conda.io/en/latest/miniconda.html). If installation
+using conda proves to be unacceptably slow, mamba can be used instead (see
+[Installation using mamba](#installation-using-mamba)).
+
+## Installation from the conda package
+
+Into a currently active, existing conda environment (>= Python 3.7)
 
     $ conda install -c conda-forge xcube
 
-Into new conda environment
+Into a new conda environment named `xcube`:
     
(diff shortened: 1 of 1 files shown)


forman (approved):
Excellent!

docs/source/installation.md
```diff
@@ -1,51 +1,95 @@
...
+[no longer recommended](https://docs.python.org/3/install/#introduction).
+Among other things, `pip` has the advantage of allowing easy deinstallation of
+installed packages.)
 
-Update
+To update the install to the latest repository version:
     
     $ activate xcube
     $ git pull --force
-    $ python setup.py develop
```
AliceBalfanz, line 53:
```suggestion
    If the packages in environment.yml have changed, please update your conda environment with: 
        $ cd xcube
        $ conda env update -n xcube --file environment.yml --prune    
```

forman: > If the packages in environment.yml have changed, please update your conda environment with: 

A user cannot know this easily!

May be better to say

"From time to time xcube's pacjage dependencis may change. If you recieve error messages about missing modules, , please update your conda environment with:"

pont-us: I'm going to add the update command to the update instructions right after the "git pull" and without any conditional, just an explanatory note that it's sometimes necessary. In most cases it will be a no-op, but executing it is easier than checking whether one needs to execute it :).

AliceBalfanz (approved):

AliceBalfanz:
Thanks for the update! :)

Issue #370, opened 2020-11-18
Let "xcube io dump" use store configuration


forman:
**Is your feature request related to a problem? Please describe.**

The CLI command `xcube io dump` does not use any store configuration. It instantiates stores with default settings.

**Describe the solution you'd like**

Add an option `-c/--config` that specifies a JSON file with configured data store instances. This should make use of the new feature provided by #364.

PR #372, opened 2020-11-18
Forman 370 io dump with config

forman:
CLI tool `xcube io dump` now has new `--config` and `--type` options. (#370)

Checklist:

### DIFF
--- a/test/core/store/stores/test_s3.py
+++ b/test/core/store/stores/test_s3.py
@@ -6,8 +6,12 @@ import xarray as xr
 from test.s3test import S3Test, MOTO_SERVER_ENDPOINT_URL
 from xcube.core.new import new_cube
 from xcube.core.store import DataStoreError
+from xcube.core.store import TYPE_SPECIFIER_CUBE
+from xcube.core.store import TYPE_SPECIFIER_DATASET
 from xcube.core.store import new_data_store
 from xcube.core.store.stores.s3 import S3DataStore
+from xcube.util.jsonschema import JsonObjectSchema
+
 
 BUCKET_NAME = 'xcube-test'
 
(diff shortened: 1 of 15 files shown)


dzelge (changes requested):
When xcube attempts loading an unknown store, the error message should contain an info that the corresponding plugin could be missing. Could because a missing plugin is not the only error that can happen. 

https://github.com/dcs4cop/xcube/blob/e9819fbfdc566b8ff92b5e926dff23b42eef65f5/xcube/core/store/store.py#L70

forman:
> When xcube attempts loading an unknown store, the error message should contain an info that the corresponding plugin could be missing. Could because a missing plugin is not the only error that can happen.
> 
> https://github.com/dcs4cop/xcube/blob/e9819fbfdc566b8ff92b5e926dff23b42eef65f5/xcube/core/store/store.py#L70

Right, I'll do!

dzelge:
@forman One thing I realies is that the title/description is not set correctly or not correctly picked up by the xcube-gen-ui. The drop down list it basically correct, but the store's title is not listed:

image

forman:
> @forman One thing I realies is that the title/description is not set correctly or not correctly picked up by the xcube-gen-ui. The drop down list it basically correct, but the store's title is not listed:

Fixed.

forman:
> ...drop down list it basically correct, but the store's title is not listed:

I'll also update the UI to show the ID, if title is missing

test/core/store/stores/test_s3.py
```diff
@@ -62,6 +66,29 @@ def test_get_open_data_params_schema(self):
...
+        self.assertIsInstance(schema, JsonObjectSchema)
+        self.assertEqual({}, schema.properties)
+        self.assertEqual(False, schema.additional_properties)
+
+        schema = self.store.get_search_params_schema(type_specifier='geodataframe')
+        self.assertIsInstance(schema, JsonObjectSchema)
+        self.assertEqual({}, schema.properties)
+        self.assertEqual(False, schema.additional_properties)
+
+    # def test_search_data(self):
```
TonioF, line 80: If you commit code in comments, please add why this is in a comment.

EDIT: Maybe add a test like test_write_and_read_delete? Or create a new test class for the two methods?

xcube/core/store/stores/memory.py
```diff
@@ -65,7 +66,8 @@ def get_type_specifiers_for_data(self, data_id: str) -> Tuple[str, ...]:
         type_specifier = get_type_specifier(self._data_dict[data_id])
         return str(type_specifier),
 
-    def get_data_ids(self, type_specifier: str = None, include_titles: bool = True) -> Iterator[Tuple[str, Optional[str]]]:
+    def get_data_ids(self, type_specifier: str = None, include_titles: bool = True) -> Iterator[
```
pont-us, line 69: Strange line break -- typo?

pont-us (approved):

xcube/cli/io.py
```diff
@@ -172,38 +174,79 @@ def writer_info(writer_id: str):
...
+              help='Store configuration filename. May use JSON or YAML format.')
+@click.option('-t', '--type', 'type_specifier', metavar='TYPE',
+              help='Type specifier. If given, only data resources that satisfy the '
+                   'type specifier are listed. E.g. "dataset" or "dataset[cube]"')
+def dump(output_file_path: str, config_file_path: Optional[str], type_specifier: Optional[str]):
     """
-    Dump all store metadata.
-    Dumps all store metadata and their dataset descriptors into a single JSON file.
+    Dump metadata of given data stores.
+    Dumps metadata for every data resources for given data stores.
```
TonioF, line 193: Alternative Suggestion:
"Dump metadata of data store resources
Dumps metadata of every data resource of available data stores.
A restriction on which data stores are considered may be passed via the CONFIG option."

If you don't approve, please at least correct "for every data resources".

TonioF (commented):
Two comments, none on the code. Looks good!

test/core/store/stores/test_s3.py
forman: > If you commit code in comments, please add why this is in a comment.

I'll do.

> EDIT: Maybe add a test like test_write_and_read_delete? Or create a new test class for the two methods?

In general I agree. But I don't have the time now. It is an implementation helper and as such its code is covered multiple times by mem, dir, and s3 store tests.

xcube/cli/io.py
forman: > If you don't approve, please at least correct "for every data resources".

I believe "for every data resources" is no good English.
I'll try to improve docs.

Issue #377, opened 2020-11-23
Determine correct filename extension


TonioF:
If you use the directory store to write a dataset (store.write_data(dataset)), it will be saved with a '.levels'-extension. This is because the filename extension is incorrectly determined.

PR #378, opened 2020-11-23
Toniof 377 determine correct extension

TonioF:
Solves https://github.com/dcs4cop/xcube/issues/377. We really need to watch closely whether to use satisfies or is_satisfied_by .


forman (approved):
Thanks!

Issue #392, opened 2021-01-06
Normalisation may be slow and consume much memory


TonioF:
In cases where the longitude of a dataset is in the range from 0 to 360 degrees, a normalization routine may be employed to relocate the values to lie between -180 and +180 degrees. The current implementation reads all values from all values from all data sets and rearranges them. In many cases, this behaviour is very slow and expensive in terms of memory.

PR #393, opened 2021-01-06
Normalisation is slow

TonioF:
Solves #392 .

I have also added that a variable's encoding is updated when the dimension order is changed during the normalisation

Checklist:

~~* [ ] Add docstrings and API docs for any new/modified user-facing classes and functions~~
~~* [ ] New/modified features documented in `docs/source/*`~~

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@
   with observe_dask_progress('Writing dataset', 100):
       dataset.to_zarr(store)  
   ```
+* Normalisation may be slow and consume much memory. (#392)
 
 ## Changes in 0.6.1
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,7 +7,10 @@
   with observe_dask_progress('Writing dataset', 100):
       dataset.to_zarr(store)  
   ```
-* Normalisation may be slow and consume much memory. (#392)
+* The xcube normalisation process, which ensures that a dataset meets the requirements 
+  of a cube, internally requested a lot of data, causing the process to be slow and
+  expensive in terms of memory consumption. This problem was resolved by avoiding to
+  read in these large amounts of data. (#392)
 
 ## Changes in 0.6.1
 
--- a/xcube/core/normalize.py
+++ b/xcube/core/normalize.py
@@ -722,6 +722,8 @@ def _normalize_dim_order(ds: xr.Dataset) -> xr.Dataset:
                 ds = ds.copy()
                 copy_created = True
             ds[var_name] = var.transpose(*dim_names)
+            if var.encoding:
+                ds[var_name].encoding['chunks'] = ds[var_name].data.chunksize
 
     return ds
 
--- a/xcube/core/normalize.py
+++ b/xcube/core/normalize.py
@@ -201,12 +201,8 @@ def _normalize_lon_360(ds: xr.Dataset) -> xr.Dataset:
     new_vars = dict()
     for var_name in var_names:
         var = ds[var_name]
-        if len(var.dims) >= 1 and var.dims[-1] == 'lon':
-            values = np.copy(var.values)
-            temp = np.copy(values[..., : lon_size_05])
-            values[..., : lon_size_05] = values[..., lon_size_05:]
-            values[..., lon_size_05:] = temp
-            new_var = xr.DataArray(values, dims=var.dims, attrs=var.attrs)
+        if 'lon' in var.dims:
+            new_var = var.roll(lon=lon_size_05, roll_coords=False)
             new_var.encoding.update(var.encoding)
             new_vars[var_name] = new_var
 
--- a/xcube/core/tile.py
+++ b/xcube/core/tile.py
@@ -102,8 +102,16 @@ def get_ml_dataset_tile(ml_dataset: MultiLevelDataset,
     return tile
 
 
-def new_rgb_image(ml_dataset, image_id, var_names, norm_ranges, z, labels,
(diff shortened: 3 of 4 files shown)


TonioF:
Tests already exist.
Second and third bullet do not apply.

CHANGES.md
```diff
@@ -7,6 +7,7 @@
   with observe_dask_progress('Writing dataset', 100):
       dataset.to_zarr(store)  
   ```
+* Normalisation may be slow and consume much memory. (#392)
```
forman, line 10: That's just the symptom and not the solution. Please be clear about what has changed in xcube, so users can undestand what's new in the upcoming release.

xcube/core/normalize.py
```diff
@@ -201,12 +201,8 @@ def _normalize_lon_360(ds: xr.Dataset) -> xr.Dataset:
...
     for var_name in var_names:
         var = ds[var_name]
-        if len(var.dims) >= 1 and var.dims[-1] == 'lon':
-            values = np.copy(var.values)
-            temp = np.copy(values[..., : lon_size_05])
-            values[..., : lon_size_05] = values[..., lon_size_05:]
-            values[..., lon_size_05:] = temp
-            new_var = xr.DataArray(values, dims=var.dims, attrs=var.attrs)
+        if 'lon' in var.dims:
+            new_var = var.roll(lon=lon_size_05, roll_coords=False)
```
forman, line 205: Great! But why is `roll_coords=False`? Shouldn't they be rolled too?

xcube/core/normalize.py
```diff
@@ -726,6 +722,8 @@ def _normalize_dim_order(ds: xr.Dataset) -> xr.Dataset:
                 ds = ds.copy()
                 copy_created = True
             ds[var_name] = var.transpose(*dim_names)
+            if var.encoding:
```
forman, line 725: Why that condition? Anything can be in `var.encoding` so it will not be empty. On the other hand, the dictionary `var.encoding` always exists.

xcube/core/normalize.py
```diff
@@ -726,6 +722,8 @@ def _normalize_dim_order(ds: xr.Dataset) -> xr.Dataset:
...
+                ds[var_name].encoding['chunks'] = ds[var_name].data.chunksize
```
forman, line 726: `ds[var_name].data.chunksize` requires that `ds[var_name].data` has an attribute `chunksize` but that cannot be assumed. It is safer to use the xarray API, but only if it is chunked:
```suggestion
            if 'chunks' in var.encoding and ds[var_name].chunks is not None:
                ds[var_name].encoding['chunks'] = ds[var_name].chunks
```

forman: Even this might lead to problems because dask chunking is more general than Zarr chunking. So there are dask chunks that cann not be used in Zarr. Zarr requires all but the last chunk to be same size, while dask allows for any chunk sizes greater zero.

forman (changes requested):
Good fix! But there are issues still...

forman:
> Tests already exist.
> Second and third check box do not apply.

Please use strikethrough to indicate what does not apply. If tests exists, place check.

xcube/core/normalize.py
TonioF: No, they shouldn't. We just took the first half of the lon coordinates and put them after the second half. The purpose of the roll command is that the variables are aligned with the correct longitude values again. If we would roll the coordinates, too, the relationship between the data and the dimension would not change and stay incorrect (see test test_fix_360_lon in test_normalize . It would fail if you'd set roll_coords to True).

xcube/core/normalize.py
TonioF: I'd like to use 

if 'chunksize' in ds[var_name].data: 

then. The reasong behind this change is that ds[var_name].encoding['chunks'] and ds[var_name].data.chunksize should have the same values, which is not correct anymore after the dimension order has changed.
What do you suggest?

forman: > if 'chunksize' in ds[var_name].data:

That is definitely wrong because you make the assumption that `data` implements `__contains__()`, which you can't.

CHANGES.md
```diff
@@ -7,6 +7,7 @@
   with observe_dask_progress('Writing dataset', 100):
       dataset.to_zarr(store)  
   ```
+* Avoid accessing variable data during normalisation. (#392)
```
forman, line 10: Please tell the user what you mean by "normalisation" and why the change was made.

xcube/core/normalize.py
forman: > which is not correct anymore after the dimension order has changed.

I don't get that. In this case the order of `ds[var_name].chunks` changed too. So why is there a problem?

xcube/core/normalize.py
forman: Condition is still wrong.

forman (changes requested):

xcube/core/normalize.py
TonioF: We are still discussing what to put there instead.

xcube/core/normalize.py
TonioF: It is required during the determination of the zarr chunks that any given chunking in the encoding corresponds to the chunking of the data sets. See here: https://github.com/pydata/xarray/blob/5735e163bea43ec9bc3c2e640fbf25a1d4a9d0c0/xarray/backends/zarr.py#L159

forman (approved):
Ok!

PR #398, opened 2021-01-19
Toniof xcube397 threadsafe progress monitors

TonioF:
There is no central progress state stack anymore. Instead, every progress state knows its parent state. A state may be the parent of more than one other state. To determine the parent, the traceback is considered. This works for progresses happening within the same thread, including tasks started with asyncio. Progresses started from another thread will have no parent (unless there is already another progress from the same thread).

Related issue: #397 

Checklist:

* [x] Add unit tests and/or doctests in docstrings
~~* Add docstrings and API docs for any new/modified user-facing classes and functions~~
~~*New/modified features documented in `docs/source/*~~
* [x] Changes documented in `docs/CHANGES.md`
* [ ] AppVeyor and Travis CI passes
* [ ] Test coverage remains or increases (target 100%)
* [ ] Associated issues closed after merge

PR #399, opened 2021-01-19
Toniof xcube397 threadsafe progress monitors dask

TonioF:
This is a follow-up pr to #398 . In this pr, parent states might be detected across threads. The principle is that dask allows to get the children threads of a thread. We can save them to a state, and when we look for a parent state, we can link a progress state from one thread to a parent progress state from another thread.
Issues with this:
- It forces us to check whether a private attribute '_children' exists for a thread, and, if not, to set it. This is because this attribute is not set for every thread that has a thread pool, but only for the one that first loads the module's __init__.py.
- It is not compatible with observe_dask_progress. There might be instances where it cannot be decided whether observe_dask_progress or observe_nested_dask_progress is the correct choice. We could try to harmonize, though.

Checklist:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,14 @@
   with observe_dask_progress('Writing dataset', 100):
       dataset.to_zarr(store)  
   ```
+* xcube progress monitoring can deal with multiple nested progress monitors. Progress from different threads are handled separately.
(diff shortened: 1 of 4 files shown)


forman (commented):
Same issue as with #398.

TonioF:
Does not work anymore. Also, it was not a promising approach to begin with.

PR #402, opened 2021-01-28
Toniof enhance dataset descriptor

TonioF:
This PR contains enhancements for the DatasetDescriptor. It is now allowed to add a dataset's coords to the descriptor. Also, the data variables are not provided in a list but as a dictionary to facilitate the lookup of a variable.

Checklist:

~[ ] Associated issues closed after merge~


TonioF:
Note: This PR is likely to cause incompatibilities with xcube-sh, xcube-cci and xcube-cds.

CHANGES.md
```diff
@@ -1,5 +1,7 @@
 ## Changes in 0.6.2 (in development)
 
+* Added `coords` parameter to DatasetDescriptor. 
+  Also, show a DatasetDescriptor' data variables as dictionary.
```
forman, line 4:
```suggestion
* Added `coords` property to `DatasetDescriptor` class. 
  The `data_vars` property of the `DatasetDescriptor` class is now dictionary. 
```

test/core/store/test_descriptor.py
```diff
@@ -114,16 +138,17 @@ def test_to_dict(self):
...
-                                         ],
+                              data_vars=dict(xf=dict(name='xf',
+                                                     dtype='rj',
+                                                     dims=('dfjhrt', 'sg'),
+                                                     ndim=2,
+                                                     attrs=dict(ssd=4,
+                                                                zjgrhgu='dgfrf'
+                                                                )
+                                                     )
+                                             ),
```
forman, line 151: Please use agreed coding style.

forman (changes requested):

TonioF:
Both requested changes applied.

test/core/store/test_descriptor.py
```diff
@@ -75,13 +76,35 @@ def test_from_dict_full(self):
...
+                               data_vars=dict(xf=dict(name='xf',
+                                                      dtype='rj',
+                                                      dims=('dfjhrt', 'sg'),
+                                                      ndim=2,
+                                                      attrs=dict(ssd=4,
+                                                                 zjgrhgu='dgfrf'
+                                                                 )
+                                                      )
+                                              )
+                               )
```
forman, line 99: The coding style issue remains.

xcube/core/store/descriptor.py
```diff
@@ -119,6 +129,22 @@ class DatasetDescriptor(DataDescriptor):
...
+    :param crs: A coordinate reference system identifier, as an EPSG, PROJ or WKT string
+    :param bbox: A bounding box of the data (order: min lon, min lat, max lon, max lat)
+    :param spatial_res: The spatial resolution of the data set
+    :param time_range: Start and end time delimiting this data's temporal extent
+    :param time_period: The data's periodicity if it is evenly temporally resolved.
+    :param coords: A list of the dataset's data coordinates
+    :param dims: A mapping of the dataset's dimensions to their sizes
+    :param data_vars: A mapping of the dataset's variable names to VariableDescriptors
+    (``xcube.core.store.VariableDescriptor``). For backwards compatibility, also lists of
+    VariableDescriptors are accepted. This compatibility will come to an end with a future release.
```
forman, line 144: I'd vote for breaking compatibility now, The framework is new and next release will be a `0.7` anyway. We unfortunatley do not have enough resources to maintain this kind of deprecations.

xcube/core/store/descriptor.py
```diff
@@ -129,8 +155,10 @@ def __init__(self,
                  spatial_res: float = None,
                  time_range: Tuple[Optional[str], Optional[str]] = None,
                  time_period: str = None,
+                 coords: Sequence[str] = None,
                  dims: Mapping[str, int] = None,
-                 data_vars: Sequence['VariableDescriptor'] = None,
+                 data_vars: Union[Sequence['VariableDescriptor'],
+                                  Mapping[str, 'VariableDescriptor']] = None,
```
forman, line 161: Get rid of `Sequence`.

```suggestion
                 data_vars: Mapping[str, 'VariableDescriptor'] = None,
```

xcube/core/store/descriptor.py
```diff
@@ -141,36 +169,55 @@ def __init__(self,
                          time_range=time_range,
                          time_period=time_period,
                          open_params_schema=open_params_schema)
+        self.coords = coords
         self.dims = dict(dims) if dims else None
-        self.data_vars = list(data_vars) if data_vars else None
+        if data_vars:
+            if type(data_vars) == dict:
```
forman, line 175: This is wrong w.r.t. type declaration. Fix:

```suggestion
            if instanceof(data_vars, Mapping):
```

xcube/core/store/descriptor.py
```diff
@@ -141,36 +169,55 @@ def __init__(self,
...
+                self.data_vars = data_vars
+            else:
+                self.data_vars = {}
```
forman, line 178: If it is really required to maintain compatibility, issue a [deprecation warning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) here.

xcube/core/store/descriptor.py
```diff
@@ -141,36 +169,55 @@ def __init__(self,
...
         assert_in('data_id', d)
-        data_vars = []
-        data_vars_as_dicts = d.get('data_vars', [])
-        for data_var_as_dict in data_vars_as_dicts:
-            data_vars.append(VariableDescriptor.from_dict(data_var_as_dict))
+        data_vars_dict = None
+        data_vars = d.get('data_vars', None)
+        if data_vars:
+            data_vars_dict = {}
+            if type(data_vars) == dict:
```
forman, line 193: Insufficient condition. 
```suggestion
            if instanceof(data_vars, Mapping):
```

xcube/core/store/descriptor.py
```diff
@@ -141,36 +169,55 @@ def __init__(self,
...
+                for data_var_key, data_var_value in data_vars.items():
+                    data_vars_dict[data_var_key] = VariableDescriptor.from_dict(data_var_value)
+            else:
+                for data_var_as_dict in data_vars:
```
forman, line 197: Get rid of supporting lists.

xcube/core/store/descriptor.py
```diff
@@ -141,36 +169,55 @@ def __init__(self,
...
-        self.data_vars = list(data_vars) if data_vars else None
+        if data_vars:
+            if type(data_vars) == dict:
+                self.data_vars = data_vars
+            else:
+                self.data_vars = {}
+                for data_var in data_vars:
+                    self.data_vars[data_var.name] = data_var
+        else:
+            self.data_vars = None
```
forman, line 182: Overall: too complex logic for simple `data_vars` assignment here. If we maintain compatibility at all, just normalize argument and assign to `self.data_vars` which is then always a dict or none.

forman (changes requested):

xcube/core/store/descriptor.py
TonioF: Okay with me. Waiting for @pont-us opinion.

xcube/core/store/descriptor.py
```diff
@@ -119,6 +129,22 @@ class DatasetDescriptor(DataDescriptor):
...
     A descriptor for a gridded, N-dimensional dataset represented by xarray.Dataset.
     Comprises a description of the data variables contained in the dataset.
+
+    :param data_id: An identifier for the data
+    :param type_specifier: A type specifier for the data
+    :param crs: A coordinate reference system identifier, as an EPSG, PROJ or WKT string
+    :param bbox: A bounding box of the data (order: min lon, min lat, max lon, max lat)
+    :param spatial_res: The spatial resolution of the data set
+    :param time_range: Start and end time delimiting this data's temporal extent
+    :param time_period: The data's periodicity if it is evenly temporally resolved.
```
pont-us, line 139: The time specification conventions are probably too lengthy to include in full here, but a "See data store conventions document" note might be useful.

xcube/core/store/descriptor.py
```diff
@@ -65,6 +65,16 @@ class DataDescriptor:
     """
     A generic descriptor for any data.
     Also serves as a base class for more specific data descriptors.
+
```
pont-us, line 68: Some of my comments on the DatasetDescriptor docstrings also apply to the corresponding parameters here, but it wasn't worth duplicating them.

xcube/core/store/descriptor.py
```diff
@@ -181,7 +228,13 @@ def _get_base_type_specifier(cls) -> TypeSpecifier:
 class VariableDescriptor:
     """
     A descriptor for dataset variable represented by xarray.DataArray instances.
-    They are part of dataset descriptor for an gridded, N-dimensional dataset represented by xarray.Dataset.
+    They are part of dataset descriptor for an gridded, N-dimensional dataset represented by
+    xarray.Dataset.
+
+    :param name: The variable name
+    :param dtype: The data type of the variable.
```
pont-us, line 235: I *think* that we're using `numpy.sctypeDict` format for type definitions, but would be glad of confirmation from you and @forman . Whatever format we're using, it should be specified here.

```suggestion
    :param dtype: The data type of the variable. Must be a string from `numpy.sctypeDict.keys()`
```

pont-us (changes requested):

xcube/core/store/descriptor.py
TonioF: As our software just reads and passes on this bit of information, I think it is okay if we don't make any assumptions about this. It may probably even be better, as this allows for different data stores to interpret this differently.

TonioF:
Next version. I have removed any support for data variable lists completely and updated the doc strings according to @pont-us suggestions. While doing this, I realized that the property spatial_res only makes sense for gridded data, so I removed it from the datadescriptor class.

xcube/core/store/descriptor.py
pont-us: Fair enough, makes sense -- but I think that that should also be made explicit with a note along the lines of "Syntax of data type specifiers is defined by the data store.".

pont-us (changes requested):
Only one additional change suggested: https://github.com/dcs4cop/xcube/pull/402#discussion_r572124387

forman (approved):
Ok!

PR #409, opened 2021-02-15
"xcube gen2" refactorings, part I

forman:
The "xcube gen2" tool has been refactored towards cleaner design. In this first round functions have been converted to classes so that the code becomes more readable, extendible, and easier to test. New ABCs `CubeProcessor` and `CubesProcessor` have been introduced.

Checklist:

* [ ] ~Changes documented in `docs/CHANGES.md`~


xcube/core/gen2/error.py
```diff
@@ -20,6 +20,6 @@
 # SOFTWARE.
 
 
-class GenError(Exception):
+class CubeGeneratorError(ValueError):
```
TonioF, line 23: Can we say with certainty that every CubeGeneratorError is a ValueError?

xcube/core/gen2/generator.py
```diff
@@ -0,0 +1,128 @@
...
+        if store_pool is not None:
+            assert_instance(store_pool, DataStorePool, 'store_pool')
+
+        self._gen_config = gen_config
+        self._store_pool = store_pool if store_pool is not None \
+            else DataStorePool()
+        self._verbose = verbose
+
+    @classmethod
+    def from_file(cls,
```
TonioF, line 64: As you can pass more than one file as parameter: Rename to from_files or, what I prefer, from_config

xcube/core/gen2/opener.py
```diff
@@ -0,0 +1,98 @@
...
+            opener = store
+            open_params.update(opener_id=opener_id, **open_params)
+        else:
+            opener = new_data_opener(opener_id)
+            open_params.update(**store_params, **open_params)
+        open_params_schema = opener.get_open_data_params_schema(input_config.data_id)
+        cube_params = {k: v for k, v in all_cube_params.items() if k in open_params_schema.properties}
+        cube = opener.open_data(input_config.data_id, **open_params, **cube_params)
+        if normalisation_required:
+            cube = normalize_dataset(cube)
```
TonioF, line 97: Actually, we cannot be certain that the dataset is a cube after the normalisation. I think it would be better to use verify_cube here and return nothing in case the verification fails.
I realize this issue existed before your PR and therefore is not really part of it, but I just happen to notice it whilst I was reviewing your changes.

TonioF (changes requested):
Overall good. I like the separation between merging and resampling, but I have a few minor questions and suggestions.

xcube/core/gen2/opener.py
forman: This is your code.

xcube/core/gen2/generator.py
forman: No, it is the one file positional arg, the other file params are keywork arguments, hence options.

xcube/core/gen2/error.py
forman: It is common to use Value Error as base class for derived errors as ValueError is generic (see zarr, xarray and others). It helps handling errors, if a derived error is one of the base errors.

TonioF: ok

xcube/core/gen2/opener.py
TonioF: I acknowledge that. What do you suggest we should do with the issue, though?
1. Ignore?
2. Handle here?
3. Handle elsewhere?

xcube/core/gen2/generator.py
TonioF: In this case the first sentence of the docstring is a bit misleading. I still prefer "config", even more so as the configuration may be piped.

xcube/core/gen2/opener.py
forman: > Actually, we cannot be certain that the dataset is a cube after the normalisation.

Why not? Normalisation may just fail or do nothing at all.

xcube/core/gen2/generator.py
forman: `form_config` doesn't make sense, because it is always constructed from configuration. The difference here is that configuration comes from a _file_ rather than a string, pipe stream, HTTP, whatsoever.

xcube/core/gen2/opener.py
forman: Therefore:

1. Ignore for time being until we face a new blocker.

PR #410, opened 2021-02-17
Toniof s3 store

TonioF:
Added missing implementations of s3 data store methods:
- register_data
- deregister_data
- describe_data
Data is registered in a json file at the base of the bucket, if no such file exists, the registry exists purely in memory.
To describe zarr datasets, two new classes S3ZarrDescriber and DirectoryZarrDescriber have been added. The describe_data-method of the s3 data store will use the S3ZarrDescriber in case no entry is found in the registry.

Checklist:

* [X] Add unit tests and/or doctests in docstrings
* [X] Add docstrings and API docs for any new/modified user-facing classes and functions
~~ [ ]New/modified features documented in `docs/source/*`~~
* [X] Changes documented in `docs/CHANGES.md`
~~[ ] Associated issues closed after merge~~


CHANGES.md
```diff
@@ -1,5 +1,7 @@
 ## Changes in 0.6.2 (in development)
 
+* Finished implementation of s3 data store
```
forman, line 3: Please write change logs by using the perspective of an interested user. So what is the reason and benefit of the change?

xcube/core/store/descriptor.py
```diff
@@ -49,7 +49,31 @@
...
+        if 'time_coverage_start' in data.attrs:
+            time_coverage_start = data.time_coverage_start
+        if 'time_coverage_end' in data.attrs:
+            time_coverage_end = data.time_coverage_end
+        return DatasetDescriptor(data_id=data_id,
+                                 type_specifier=get_type_specifier(data),
+                                 bbox=bbox,
+                                 time_range=(time_coverage_start, time_coverage_end),
+                                 dims=dict(data.dims),
+                                 data_vars=variable_descriptors)
```
forman, line 76: We require also `spatial_res` (an optional `Tuple[float, float]`) and `time_period` (an optional `str`).

xcube/core/store/stores/s3.py
```diff
@@ -59,13 +61,16 @@
     '.levels': (TYPE_SPECIFIER_MULTILEVEL_DATASET, 'levels', _STORAGE_ID),
 }
 
+_REGISTRY_FILE = 'registry.json'
```
forman, line 64: What is the benefit of another JSON file? In Zarr, all metadata is already in JSON format.

xcube/core/store/zarrdescriber.py
```diff
@@ -0,0 +1,183 @@
...
+    """
+
+    def __init__(self, path: str):
+        self._path = path
+
+    def _open_file(self, data_id: str, file: str):
+        return open(f'{self._path}/{data_id}/{file}', 'r')
+
+    def _list_zarr_dir_content(self, data_id: str):
+        return os.listdir(f'{self._path}/{data_id}')
```
forman, line 183: Move this into a separate implementation package, e.g. into `accessors` or a new one. As an implementation detail, it does not belong to the data store framework.

For the future, I'd prefer, if we model a data _registry_ or data _index_ instead, so 

- we are no longer bound to files (_open_file, _list_zarr_dir_content);
- the registry may be backed by an efficient cloud indexing software, e.g. Apache Solr, MongoDB, REDIS, ...

forman (changes requested):
Changes may be best discussed in a TC.

xcube/core/store/zarrdescriber.py
TonioF: Do you mean the whole module? I'd agree. I think a new package 'describers' would be appropriate. Okay?

xcube/core/store/descriptor.py
TonioF: I agree, but these two are a little harder to determine, as they are not given as metadata, but must be determined from the data. I'll go ahead.

CHANGES.md
```diff
@@ -1,5 +1,8 @@
 ## Changes in 0.6.2 (in development)
 
+* S3 data store now implements describe_data.
```
forman, line 3:
```suggestion
* The S3 data store  `xcube.core.store.stores.s3.S3DataStore` now implements the `describe_data()` method. 
```

xcube/core/store/zarrdescriber.py
forman: No, this is an implementation detail that we want to hide. It does not deserve an own package. Move this module as a private one into `accessors` or `stores` and don't let its classes escape into the public store API.

forman (changes requested):
Overall very good, but hide describers from public api and move it as an impl. detail into one of the implementation subpackages `accessors` or `stores`.

forman (approved):
Supi!

PR #411, opened 2021-02-17
xcube generator service CLI and Python API

forman:
The `xcube gen2` tool has new options so users can now 

- submit requests to a remote xcube generator service using the `--service` option;
- pass `--info` to dump information about the cube to be generated.

Checklist:


CHANGES.md
```diff
@@ -1,5 +1,10 @@
 ## Changes in 0.6.2 (in development)
 
+* The `xcube gen2` data cube generator tool has been hidden from
```
AliceBalfanz, line 3: When I have this branch checked out and call `xcube --help` I would expect `gen2` not to appear in the list because it is supposed to be hidden. I have done `python setup.py develop` beforehand. 

image

forman: It _does_ not appear. It is a problem of your env.

PR #419, opened 2021-03-09
Workaround for some cases that end up in #347

### LINKED ISSUE #347: Writing to zarr fails with message "specified zarr chunks would overlap multiple dask chunks"
For some data it is necessary to rechunk it in order to deal with it efficiently. This rechunking process might result in that the final chunk of a dimension has a smaller size than the previous ones. E.g., if we have a dimension of 500, originally split into ten chunks of size 50, we can rechunk it to sizes of 200, 200, and 100. This is perfectly valid and supported by xarray.
However, when the affected dimension/variable is latitude and the latitude is ascending, the current implementation of the normalization in xcube will revert the order of chunks and cause that the smaller chunk will be at the start - which is not supported by xarray and might result in an error as the one in the description, for example when writing to zarr.

For the time being, I will work around this issue by ensuring that all chunks are of the same size after rechunking, which will make this issue a bit harder to reproduce. I open this issue to document that the current solution of dealing with ascending latitudes is not optimal. The best solution will probably be to get rid of the part where latitudes are reverted during the normalization and support ascending latitudes in the data.

See also https://github.com/dcs4cop/xcube/issues/251 and https://github.com/dcs4cop/xcube/issues/327.

forman:
* Dataset normalisation no longer includes reordering increasing
  latitude coordinates, as this creates datasets that are no longer writable 
  to Zarr. (addresses but not completely resolves #347)

Checklist:

* [x] Test coverage remains or increases (target 100%)  -   not adressed, must remove entire method for this


test/core/test_normalize.py
```diff
@@ -136,7 +134,7 @@ def test_normalize_inverted_lat(self):
             chunks={'time': 1})
 
         actual = normalize_dataset(ds)
-        xr.testing.assert_equal(actual, expected)
+        xr.testing.assert_equal(actual, ds)
```
TonioF, line 137: The check of this test is now that latitude is not inverted anymore. One might argue that we don't need it anymore at all. At least, remove the lines were the 'expected' dataset is built and rename the test to 'test_lat_not_inverted'.

xcube/core/normalize.py
```diff
@@ -59,7 +59,12 @@ def normalize_dataset(ds: xr.Dataset) -> xr.Dataset:
 
     # xcube viewer currently requires decreasing latitude co-ordinates, so
     # we invert them here if necessary.
-    ds = _ensure_lat_decreasing(ds)
+
+    # TODO: commented out as _ensure_lat_decreasing() produces
+    #   chunks in "lat" that have the smallest chunk first. This will
+    #   fail when writing to Zarr. (Hack for #347)
+    #   See also https://github.com/pydata/xarray/issues/2300
+    # ds = _ensure_lat_decreasing(ds)
```
TonioF, line 67: Why only commented out? Why not fully removed? And the whole method along with it?

test/core/test_normalize.py
pont-us: I agree that the variable `expected` (lines 128-134) can be removed entirely, but I think it's still useful to keep the test itself as a definition of current behaviour, even if it's trivial. For the name I'd prefer `test_normalize_on_increasing_latitudes`, which just refers the input format rather than the expected behaviour, and means that there's one less place to update if/when we change it again. (I try to avoid/remove "inverted" where possible, because I find it ambiguous -- to me it just implies "the opposite of whatever we currently want in this context". "increasing" / "decreasing" are far clearer.)

xcube/core/normalize.py
pont-us: Maybe because it's only intended as a temporary hack? Not sure...

pont-us (changes requested):
Only trivial changes suggested.

Issue #420, opened 2021-03-10
DataStore.get_data_ids() to return minimum set of attributes


forman:
**Is your feature request related to a problem? Please describe.**

So far, all clients of the xcube data store framework require at least a dataset title in addition to the sole dataset ID returned by `DataStore.get_data_ids()` instance method. The CCI Toolbox now requires another attribute namely `"verification_flags"` together with the `"title"`attribute for each ID returned.

**Describe the solution you'd like**

Add keyword-argument `include_attrs: Sequence[str] = None` to `DataStore.get_data_ids()`. In case it is given, a list of tuples each comrising dataset id and attribute dictionary is returned.

**Describe alternatives you've considered**

Returning `"verification_flags"` as part of the data store metadata given in extension point registration.

**Additional context**

- [CCI Toolbox Cate](https://github.com/CCI-Tools/cate)
- https://github.com/CCI-Tools/cate-app/issues/120

PR #421, opened 2021-03-10
Addressing #420 (for CCI Toolbox)

forman:
* Slightly changed signature of `xcube.core.store.DataStore.get_dataset_ids()`
  by adding a new keyword argument `include_attrs: Sequence[str] = None` that
  can be used to obtain a minimum set of dataset attributes for each returned 
  dataset identifier. However, `include_attrs` is ignored to far in the "s3", 
  "memory", and "directory" data stores. (#420) 

Checklist:


test/core/store/stores/test_directory.py
```diff
@@ -162,23 +162,23 @@ def test_get_data_ids(self):
...
             {
-                ('cube-1-250-250.zarr', None),
-                ('cube-5-100-200.zarr', None),
-                ('cube.nc', None),
+                ('cube-1-250-250.zarr', {}),
+                ('cube-5-100-200.zarr', {}),
+                ('cube.nc', {}),
             },
-            set(self.store.get_data_ids(include_titles=False))
+            set(self.store.get_data_ids(include_attrs=["title"]))
```
TonioF, line 169: As this is not the expected behaviour, I recommend to comment this part of the test out as long as the stores do not return titles (or any other attributes).

xcube/core/store/store.py
```diff
@@ -153,22 +153,41 @@ def get_type_specifiers_for_data(self, data_id: str) -> Tuple[str, ...]:
...
-
-        The returned iterator items are 2-tuples of the form (*data_id*, *title*), where *data_id*
-        is the actual data identifier and *title* is an optional, human-readable title for the data.
-        If *include_titles* is false, the second item of the result tuple will be None.
-
-        :param type_specifier: If given, only data identifiers that are available as this type are returned. If this is
-        omitted, all available data identifiers are returned.
-        :param include_titles: If true, the store will attempt to also provide a title.
+        If a store implementation supports only a single data type, it should verify that
+        *type_specifier* is either None or compatible with the supported data type.
```
TonioF, line 165: If a store implementation supports only a single data type, it should verify that *type_specifier* is either None or compatible with the supported data type. -> A store implementation should verify that *type_specifier* is either None or compatible with a supported data type.

xcube/core/store/store.py
```diff
@@ -153,22 +153,41 @@ def get_type_specifiers_for_data(self, data_id: str) -> Tuple[str, ...]:
...
+
+        If *include_attrs* is provided, it must be a sequence of names of metadata attributes.
+        The store will then return extra metadata for each returned data resource
+        identifier according to the names of the metadata attributes as tuples (*data_id*, *attrs*).
+
+        Hence, the type of the returned iterator items depend on the value of *include_attrs*:
```
TonioF, line 171:
```suggestion
        Hence, the type of the returned iterator items depends on the value of *include_attrs*:
```

xcube/core/store/store.py
```diff
@@ -153,22 +153,41 @@ def get_type_specifiers_for_data(self, data_id: str) -> Tuple[str, ...]:
...
+
+        - If *include_attrs* is None (the default), the method returns an iterator
+          of dataset identifiers *data_id* of type `str`.
+        - If *include_attrs* is a sequence of attribute names, even an empty one,
+          the method returns an iterator of tuples (*data_id*, *attrs*) of type
+          `Tuple[str, Dict]`, where *attrs* is a dictionary filled according to the
+          names in *include_attrs*.
+
+        The individual attributes must not exists in the dataset's metadata, they may also be
```
TonioF, line 180: Well, the title is often part of the metadata, I don't see why the two shall be mutually exclusive. I suggest to remove this sentence.

xcube/core/store/store.py
```diff
@@ -153,22 +153,41 @@ def get_type_specifiers_for_data(self, data_id: str) -> Tuple[str, ...]:
...
+        generated on-the-fly. An example for a generic attribute name is "title".
+        A store should try to resolve ```include_attrs=["title"]``` by returning items such as
+        ```("ESACCI-L4_GHRSST-SSTdepth-OSTIA-GLOB_CDR2.1-v02.0-fv01.0.zarr",
+        {"title": "Level-4 GHRSST Analysed Sea Surface Temperature"})```.
+
+        :param type_specifier: If given, only data identifiers that are available as this type are returned.
+            If this is omitted, all available data identifiers are returned.
+        :param include_attrs: A sequence of names of attributes to be returned for each dataset identifier.
+            If given, the store will attempt to also provide the given set of dataset attributes.
```
TonioF, line 189:
```suggestion
            If given, the store will attempt to provide the set of requested dataset attributes in addition to the data ids.
```

TonioF (changes requested):
I like it, but please consider my comments.

TonioF:
One more thing: Four tests fail.

TonioF:
And, I'm sorry: Where do we put the information what attributes a store offers?

test/core/store/stores/test_directory.py
forman: This _is definitely_ the expected behaviour for _this_ implementation. I agree, adding a comment would be good. Once we intent to change behaviour, we do this by a test-driven aypproach and _change the test accordingly first_.

forman:
> And, I'm sorry: Where do we put the information what attributes a store offers?

We'll define a set of common and generic attributes that may be requested here. We put them in the method's docstring later.

Issue #423, opened 2021-03-23
xcube gen2 should create subsets from datasets returned from stores 


forman:
**Is your feature request related to a problem? Please describe.**

"xcube gen2" should create subsets from datasets returned from stores that cannot subset on their own.
Subset parameters are:

- variable_names
- bbox
- time_range

**Describe the solution you'd like**

"xcube gen2" shall detect which subset parameters are supported by a store. For the ones not supported it shall use common xarray functionality to create that subsets. 
   
**Additional context**

Urgently required by DCFS and Cate.

PR #425, opened 2021-03-23
xcube gen2 should create subsets from datasets returned from stores

forman:
`xcube gen2` now creates subsets from datasets returned by data stores that do not recognize cube subset parameters `variable_names`, `bbox`, and `time_range`. (#423)

Checklist:

* [x] Associated issues closed after merge


TonioF (approved):
The standard name for the parameter for variable names is 'variable_names', according to the store documentation. As we are not dealing with a store here, we are free to use another name, but I suggest we stay consistent across xcube interfaces.

Otherwise good!

PR #433, opened 2021-04-12
Fix for broken JSON Schema serialization

### LINKED ISSUE #432: Stores do not properly serialize dataset descriptors
**Describe the bug**

The method `DataStore.describe_data()` returns a metadata descriptor of type `DataDescriptor`. The descriptor has an optional property `open_data_params_schema` of type `JsonObjectSchema`.

When calling `DataDescriptor.to_dict()` the value of a given `open_data_params_schema` will be not be converted correctly into a dictionary:

1. simple type properties of `open_data_params_schema` will be converted by name, that is, it may not result into valid JSON Schema e.g. invalid `additional_properties` vs. valid `additionalProperties`. 
2. properties whose types are again nested JSON schema objects will not be converted at all and hence stay instances of subclasses of `JsonSchema`.

The `JsonSchema.to_dict` method is obviously not called during serialisation. But doing so, yields a infite recursion, see #431. 

**To Reproduce**

This fails:

```python
    def test_to_dict(self):
        descriptor = DatasetDescriptor(
            data_id='xyz',
            type_specifier=TypeSpecifier('dataset', flags={'cube'}),
            crs='EPSG:9346',
            bbox=(10., 20., 30., 40.),
            spatial_res=20.,
            time_range=('2017-06-05', '2017-06-27'),
            time_period='daily',
            open_params_schema=JsonObjectSchema(
                properties=dict(
                    consolidated=JsonBooleanSchema(),
                ),
                additional_properties=False,
            )
        )
        descriptor_dict = descriptor.to_dict()
        self.assertEqual(
            {
                'data_id': 'xyz',
                'crs': 'EPSG:9346',
                'type_specifier': 'dataset[cube]',
                'bbox': [10.0, 20.0, 30.0, 40.0],
                'spatial_res': 20.0,
                'time_range': ['2017-06-05', '2017-06-27'],
                'time_period': 'daily',
                'open_params_schema': {
                    'type': 'object',
                    'properties': {
                        'consolidated': {
                            'type': 'boolean'
                        }
                    },
                    'additionalProperties': False,
                },
            },
            descriptor_dict)
```

**Additional context**

xcube 0.8.0,dev7

forman:
Fixes #432 

Checklist:

* [x] Associated issues closed after merge

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 ## Changes in 0.8.0 (in development)
 
+* Fixed broken JSON object serialisation of objects returned by 
+  `DataStore.describe_object()`. (#432)
 * Changed behaviour and signature of `xcube.core.store.DataStore.get_dataset_ids()`.
   The keyword argument `include_titles: str = True` has been replaced by 
   `include_attrs: Sequence[str] = None` and the return value changes accordingly:
--- a/test/core/store/test_descriptor.py
+++ b/test/core/store/test_descriptor.py
@@ -10,6 +10,7 @@ from xcube.core.store.descriptor import GeoDataFrameDescriptor
 from xcube.core.store.descriptor import VariableDescriptor
 from xcube.core.store.descriptor import new_data_descriptor
 from xcube.core.store.typespecifier import TypeSpecifier
+from xcube.util.jsonschema import JsonBooleanSchema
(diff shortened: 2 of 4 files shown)


TonioF (approved):
One optional suggestion to simplify the if condition, otherwise fine. Tests fail, but that seems unrelated to the fixes (and I couldn't reproduce that locally).

PR #441, opened 2021-04-29
Forman xxx fix subset from large images

forman:
[Description of PR]

Checklist:


xcube/core/geocoding.py
```diff
@@ -419,7 +422,16 @@ def gu_compute_ij_bboxes(x_image: np.ndarray,
             ij_bboxes[k, 3] = j_max
 
 
-def _get_dataset_xy_names(dataset: xr.Dataset, xy_names: Tuple[str, str] = None) -> Tuple[str, str]:
+def get_xy_var(dataset: xr.Dataset, name: str) -> xr.DataArray:
```
TonioF, line 425: I'd even prefer the new name get_cood_var, but get_xy_var is also fine.

TonioF (commented):
So ... the correct solution would come with the integration of the resampling?

forman:
> So ... the correct solution would come with the integration of the resampling?

Well this one is also right, but with the resampling comes a cleaner solution.

TonioF (approved):
I

PR #466, opened 2021-06-18
use consolidated parameter to open data when possible

TonioF:
This PR ensures that consolidated is used in the s3 store's describe data method when possible. This can be a notable performance boost when the dataset is large.


forman (approved):
Good!

PR #474, opened 2021-06-29
Fix broken build and #434

### LINKED ISSUE #434: xcube.core.extract.get_cube_values_for_points() - Dataset is not a valid xcube dataset
Related to #104

The function for extracting values returns always an error

`xcube.core.extract.get_cube_values_for_points(XCube, Dictionary)`

Where: `Dictionary= dict( time='2019-03-01', lat='30.',lon='36.', method='nearest')`

The error is:
> ValueError: Dataset is not a valid xcube dataset, because:
> - coordinate variable 'lon' is not equidistant;
> - coordinate variable 'lat' is not equidistant.

What does that error mean? (or better - How should the dictionary be implemented?)

forman:
* This PR should fix the broken build due new test failures for `xcube extract --ref ...`. The actual fix is [here](https://github.com/dcs4cop/xcube/pull/474/files#diff-2a630afcb3e2c7ecff04e5868e071f9de65953c5ff4c8f32e4f1934cddb38e34R117-R122)
* Also fixes #434

FYI @AliceBalfanz @dzelge @pont-us 

Checklist:

### DIFF
--- a/xcube/core/extract.py
+++ b/xcube/core/extract.py
@@ -425,11 +425,9 @@ def _normalize_series(
         for col_name in col_names:
             if col_name in points:
                 col = points[col_name]
-                if not isinstance(col, (np.ndarray,
-                                        da.Array,
-                                        xr.DataArray,
-                                        pd.Series)):
-                    new_points[col_name] = xr.DataArray(col)
+                if not isinstance(col, xr.DataArray):
+                    col = xr.DataArray(col)
+                new_points[col_name] = col
         points = new_points
 
     num_points = None
--- a/xcube/core/extract.py
+++ b/xcube/core/extract.py
@@ -1,5 +1,7 @@
-from typing import Dict, Union, Tuple, Any, Optional, Mapping, Sequence, Hashable
+from typing import Dict, Union, Tuple, Any, Optional, \
+    Mapping, Sequence, Hashable
 
+import dask.array as da
 import numpy as np
 import pandas as pd
 import xarray as xr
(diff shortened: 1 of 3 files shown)


xcube/core/extract.py
```diff
@@ -14,89 +16,144 @@
...
+        })
+
         cube_values.update(point_refs)
 
+    # The (coordinate) variable named INDEX_DIM_NAME should not exist,
+    # otherwise cube_values.to_dataframe() will use an empty string
+    # to name that series "" in the output DataFrame instance.
+    # This seems to be an issue of xarray 0.18.2+.
+    if INDEX_DIM_NAME in cube_values:
+        cube_values = cube_values.drop_vars(INDEX_DIM_NAME)
```
forman, line 122: This fixes broken build!

xcube/core/extract.py
```diff
@@ -314,38 +408,56 @@ def get_dataset_indexes(dataset: xr.Dataset,
...
-        points = xr.Dataset(points)
-    return points
-
+def _normalize_series(
+        points: PointsLike,
+        col_names: Sequence[str],
+        force_dataset: bool = False,
+        param_name: str = "points"
+) -> Tuple[PointsLike, int]:
+    if not isinstance(points, (xr.Dataset, pd.DataFrame)):
```
TonioF, line 423: So in case the points are given as Mapping[str, SeriesLike] you will ignore them and create your own new points instead? This doesn't seem right to me. Please explain or, even better, write a test.

xcube/core/extract.py
```diff
@@ -314,38 +408,56 @@ def get_dataset_indexes(dataset: xr.Dataset,
...
+        new_points = {}
+        for col_name in col_names:
+            if col_name in points:
+                col = points[col_name]
+                if not isinstance(col, (np.ndarray,
```
TonioF, line 428: Did you forget an else here?

TonioF (changes requested):
There are two things that don't look right to me. Please check.

xcube/core/extract.py
forman: No, this is a filter. Later below col_names will be validated. This is more or less old code.

xcube/core/extract.py
TonioF: Ah, wait. You don't ignore them. Still, please see my other comment and consider adding a test.

forman: Please read #434. This code block is already covered by tests. I'll add a comment.

TonioF (approved):
Okay then.

PR #481, opened 2021-07-08
Dzelge xxx gh actions

dzelge:
When accepting this PR, xcube will have GH actions that will run unittests and build a docker image

Checklist:

* [ x ] Add unit tests and/or doctests in docstrings
* [ ] Changes documented in `docs/CHANGES.md`
* [ x ] AppVeyor and Travis CI passes
* [ x ] Test coverage remains or increases (target 100%)
* [ x ] Associated issues closed after merge


.github/workflows/xcube_workflow.yaml
```diff
@@ -0,0 +1,81 @@
...
+          verbose: true # optional (default = false)
+  build-docker-image:
+    runs-on: ubuntu-latest
+    # Only run if unittests succeed
+    needs: unittest
+    # Build the docker image and push to quay.io
+    name: build-docker-image
+    env:
+      APP_NAME: xcube
+      ORG_NAME: bcdev
```
forman, line 48: Or is it

```suggestion
      ORG_NAME: bc-org
```

CHANGES.md
```diff
@@ -15,6 +15,8 @@
 * Added a new utility module `xcube.util.temp` that allows for creating 
   temporary files and directories that will be deleted when the current 
   process ends.
+* Added a GitHub action that will unittest xcube, and build and push docker images. The
```
forman, line 18:
```suggestion
* Added a GitHub action that will run xcube unit tests, and build and push Docker images. The
```

Dockerfile
```diff
@@ -1,25 +1,21 @@
...
-FROM continuumio/miniconda3:latest
+FROM quay.io/bcdev/xcube-python-base:0.8.1
+
+ARG INSTALL_PLUGINS=1
 
 # Metadata
 LABEL maintainer="helge.dzierzon@brockmann-consult.de"
 LABEL name=xcube
-LABEL version=0.8.0.dev7
+LABEL version=0.9.0.dev0
```
forman, line 8: Would be nice if we can derive this directly from `xcube.version`.

Dockerfile
```diff
@@ -30,19 +26,26 @@ COPY environment.yml /tmp/environment.yml
...
+RUN if [[ ${INSTALL_PLUGINS} == '1' ]]; then bash install_xcube-datastore.sh xcube-cds ${XCUBE_CDS_VERSION} release; fi;
+
 # Export web server port
 EXPOSE 8080
 
 # Start bash, so we can invoke xcube CLI.
-ENTRYPOINT ["/bin/bash", "-c"]
-# By default, activate xcube environment and print usage help.
-CMD ["source activate xcube && xcube serve -v --address 0.0.0.0 --port 8080 -c ./examples/serve/demo/config.yml"]
+CMD ["/bin/bash"]
```
forman, line 51: Good!

Dockerfile
```diff
@@ -30,19 +26,26 @@ COPY environment.yml /tmp/environment.yml
...
-COPY . /xcube
+COPY . ./
 
 # Setup xcube package.
 RUN source activate xcube && python setup.py install
 
+WORKDIR /tmp
+ADD scripts/install_xcube-datastore.sh ./
+
+ENV XCUBE_SH_VERSION=0.8.0
```
forman, line 40: Can we pls move all version specifications together and top-most?

scripts/install_xcube-datastore.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/env bash
```
forman, line 1: Can we pls rename this file to just `install-xcube.sh`, if there is no special reason for `-datastore` sufix

Dockerfile
forman: But where is the activation of the xcube env?

forman (changes requested):
Great! Pls see my comments and suggestions.

Dockerfile
dzelge: Yes. That'd be nice. Could actually work in the actions.  I write that down as an issue

Dockerfile
dzelge: Yes we can. done

scripts/install_xcube-datastore.sh
dzelge: done

Dockerfile
dzelge: In this case, the user will have to do it

PR #525, opened 2021-09-13
Forman xxx gen raise for result

forman:
This PR contains numerous implementation changes in `xcube gen2`, but API changes are backward compatible:

1. Results from cube generators may now include `status_code`. If the value is set, it should be used by the generator service xcube-hub as HTTP status code. Adapted JSON schema accordingly. @dzelge Please have a look here `test/core/gen2/remote/server.py`. This code shows how eeror handlich should take place.
2. A new cube generator flag parameter `raise_on_error: bool = False`  has been introduced. If set, `CubeGeneratorError`is raised in case of error, otherwise (the default), a result object with `"status": "error"` is produced. We also use `"status_code": 400` for bad request errors and `"status_code": 500` for processing errors.
3. Progress observers are now correctly deregistered.

FYI @TonioF 

Checklist:

* [ ] AppVeyor CI passes


dzelge (approved):

dzelge (commented):
One question: There is a significant difference between the gen2 service tests and the xcube hub behaviour. The xcube test server polls when a job is created. The hub on the other hand does not do that. It immediately returns and has, hence, no result object in the response. The hub expects gen2 to poll. Does gen2 still poll?

forman:
> The xcube test server polls when a job is created.

It does no polling at all. Where have you seen that? Only the `xcube.core.gen2.remote.RemoteCubeGenerator` does polling.

PR #526, opened 2021-09-13
Fix xcube serve to skip undisplayable datasets

forman:
In server impl. introduced new exception type CubeIsNotDisplayable raised if dataset's CRS is not geographic, if x_res != y_res, or if tile_grid cannot be created. If it is raised, we reject the dataset.

Checklist:

* [x] AppVeyor CI passes


AliceBalfanz:
Something is still not quite right, I would like to show you in person, because it is a bit difficult to describe, whats wrong - but not all data is displayed, when serving the data to the viewer, and the region the viewer zooms to is quite strange.

TonioF (approved):
Good!

AliceBalfanz (approved):

Issue #516, opened 2021-09-08
Unify xcube dataset i/o


TonioF:
Currently there are parts in the xcube dataset i/o that do not use the data store framework but try to access data on their own. This leads to redundant code and an increased maintenance effort. These places should be changed so that the data store framework is used throughout and there is a single source of truth. This will result in less code, less maintenance, and a more comprehensible xcube API

- [ ] Use store framework consistently in xcube gen2
- [ ] Use store framework consistently in xcube serve
- [ ] Use store framework consistently in other xcube cli tools (gen, chunk, level, ...)
- [ ] Remove xcube.core.dsio.DatasetIO and derivates

forman:
See https://github.com/dcs4cop/xcube/issues/458

forman:
The only usage of `xcube.core.mldataset.open_ml_dataset()` seems to be the `xcube level` tool, see #617.

If we can replace this by data store API calls, we also can get rid of

* `xcube.core.mldataset.open_ml_dataset_from_object_storage()` and `xcube.core.mldataset.ObjectStorageMultiLevelDataset`
* `xcube.core.mldataset.open_ml_dataset_from_local_fs()` and `xcube.core.mldataset.FileStorageMultiLevelDataset`

PR #527, opened 2021-09-14
Toniof 516 unify datasetio

TonioF:
This PR addresses #516. It ensures that, if a Configuration with Datasets is passed to a ServiceContext, each dataset is assigned to a datastore that will be used to open the data (unless the dataset is in-memory). It is attempted to find common roots for data stores to group datasets located at same/similar locations together.

When reviewing this, please consider especially the function get_data_store_pool of the ServerContext: This function used to remove a data store pool if there were no datastores explicitly configured. I had to change this, but I don't fully understand what was the original reasoning behind this.

Checklist:

* [x] AppVeyor CI passes


xcube/core/store/fs/impl/fs.py
```diff
@@ -96,7 +96,7 @@ def get_fs_params_schema(cls) -> JsonObjectSchema:
                     default=True,
                 ),
                 requester_pays=JsonBooleanSchema(
-                    description='If "RequesterPays" buckets are supported.',
+                    description='If "RequesterPays" buckets are supporteendd.',
```
forman, line 99: What? `supporteendd` ?

xcube/core/store/storepool.py
```diff
@@ -295,6 +315,40 @@ def get_store_instance(self, store_instance_id: str) -> DataStoreInstance:
...
+    #                              'and "store_config" can be given')
+    #         self._assert_valid_instance_id(store_instance_id)
+    #         return self._instances[store_instance_id]
+    #     elif store_config is not None:
+    #         store_instance_id = self.get_store_instance_id(store_config)
+    #         return self._instances[store_instance_id]
+    #     else:
+    #         raise ValueError('Either parameter "store_instance_id" or '
+    #                          '"store_config" must be given.')
+
```
forman, line 351: Why has this been added/removed and/or commented out?

xcube/webapi/context.py
```diff
@@ -69,6 +69,10 @@
 
 # We use tilde, because it is not a reserved URI characters
 STORE_DS_ID_SEPARATOR = '~'
+FS_TYPE_TO_STORE = {
```
forman, line 72: Better `FS_TYPE_TO_STORE_ID`

xcube/webapi/context.py
```diff
@@ -899,7 +961,5 @@ def _open_ml_dataset_from_python_code(ctx: ServiceContext,
 
 
 _MULTI_LEVEL_DATASET_OPENERS = {
-    "obs": _open_ml_dataset_from_object_storage,
-    "local": _open_ml_dataset_from_local_fs,
     "memory": _open_ml_dataset_from_python_code,
 }
```
forman, line 965: Remove this one-entry dict.

xcube/webapi/context.py
```diff
@@ -491,19 +607,15 @@ def _open_ml_dataset(self, dataset_config: DatasetConfigDict) \
...
-            fs_type = dataset_config.get('FileSystem', 'local')
-            if self._ml_dataset_openers \
-                    and fs_type in self._ml_dataset_openers:
-                ml_dataset_opener = self._ml_dataset_openers[fs_type]
-            elif fs_type in _MULTI_LEVEL_DATASET_OPENERS:
-                ml_dataset_opener = _MULTI_LEVEL_DATASET_OPENERS[fs_type]
-            else:
+            fs_type = dataset_config.get('FileSystem')
+            if fs_type != 'memory':
                 raise ServiceConfigError(f"Invalid FileSystem {fs_type!r}"
```
forman, line 612: Can we come here? How does a config look like that allows coming here?

xcube/webapi/context.py
```diff
@@ -269,8 +273,110 @@ def get_dataset_configs(self) -> List[DatasetConfigDict]:
                 dataset_configs += \
                     self.get_dataset_configs_from_stores()
                 self._dataset_configs = dataset_configs
+                self._maybe_assign_store_instance_ids()
         return self._dataset_configs
 
+    def _maybe_assign_store_instance_ids(self):
```
forman, line 279: Long, complex method. Please refactor and/or add comments so one can understand the algorithm implementation. It takes me too much time to reverse-engineer the intentions coded here.

xcube/core/store/storepool.py
```diff
@@ -262,7 +262,27 @@ def store_instance_ids(self) -> List[str]:
     def store_configs(self) -> List[DataStoreConfig]:
         return [v.store_config for k, v in self._instances.items()]
 
-    def has_store_config(self, store_instance_id: str) -> bool:
+    def get_store_instance_id(self,
+                              store_config: DataStoreConfig,
+                              strict_check: bool = False
+                              ) \
+            -> Optional[str]:
```
forman, line 269: Code style:

```suggestion
    def get_store_instance_id(self,
                              store_config: DataStoreConfig,
                              strict_check: bool = False
                              ) \
            -> Optional[str]:
```
```suggestion
    def get_store_instance_id(
        self,
        store_config: DataStoreConfig,
        strict_check: bool = False
) -> Optional[str]:
```

forman: GitHub has a bug here: it doubles the suggestions! Only the second is valid!

forman: It even doubles every comment!

forman (changes requested):
I believe, this is very good!

However I have some trouble understanding the base algorithm implemented here. The code urgently needs some explanation, may it be docstrings, comments, or by refactorings. Please have a look at the comments.

xcube/core/store/fs/impl/fs.py
TonioF: Not only accidentally commited, accidentally changed. Seems I wrote 'end' when the wrong window had the focus. I change it back.

xcube/core/store/storepool.py
TonioF: This was an early implementation to determine stores based on a storeconfig. I found that the newly added method get_store_instance_id was enough, so I commented these out but forgot to remove them. Removing them now.

xcube/webapi/context.py
TonioF: Agree.

xcube/webapi/context.py
TonioF: done

xcube/webapi/context.py
TonioF: That would be a config that has a filesystem that is neither local nor obs nor memory, as the error message says. I just checked and found that we would already get an error earlier, as it would be attempted to create a data store from an unknown file system. That error would break the service context and prevent that any dataset can be opened. I think it would be better to change it back so that this error comes up that is specific to one data set id. Do you agree?

xcube/core/store/storepool.py
TonioF: I just wanted to check our code style for this issue. I did not find it, I just recall we already talked about this before.

But are you certain that the suggestion below is right? Going to the lowest indention level with the method header?

xcube/webapi/context.py
TonioF: Okay, I tried.

xcube/webapi/context.py
forman: Yes.

xcube/core/store/storepool.py
forman: > I did not find it, I just recall we already talked about this before.

PEP-8. And yes, we talked about this before, see https://www.python.org/dev/peps/pep-0008/#indentation

> Going to the lowest indention level with the method header?

No that's not right, it happened per accident.

xcube/webapi/context.py
forman: `FS_TYPE_TO_PROTOCOL` then. See #528

forman (commented):
Good, let's talk about `_maybe_assign_store_instance_ids(self)` tomorrow!

xcube/webapi/context.py
TonioF: Okay!

xcube/core/store/storepool.py
TonioF: Okay, changed it according to pep8

PR #534, opened 2021-09-17
Versions follow-up

forman:
* Added CLI "xcube versions"
* Adjusted usage of `get_xcube_versions()`
* `get_xcube_versions()` returns chached instance

[Description of PR]

Checklist:

* [ ] AppVeyor CI passes

### DIFF
--- a/xcube/cli/versions.py
+++ b/xcube/cli/versions.py
@@ -1,5 +1,5 @@
 # The MIT License (MIT)
-# Copyright (c) 2019 by the xcube development team and contributors
+# Copyright (c) 2021 by the xcube development team and contributors
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy of
 # this software and associated documentation files (the "Software"), to deal in
--- a/xcube/cli/versions.py
+++ b/xcube/cli/versions.py
@@ -1,0 +1,70 @@
+# The MIT License (MIT)
+# Copyright (c) 2019 by the xcube development team and contributors
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy of
+# this software and associated documentation files (the "Software"), to deal in
+# the Software without restriction, including without limitation the rights to
+# use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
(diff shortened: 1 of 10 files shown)


xcube/cli/versions.py
```diff
@@ -0,0 +1,70 @@
+# The MIT License (MIT)
+# Copyright (c) 2019 by the xcube development team and contributors
```
TonioF, line 2:
```suggestion
# Copyright (c) 2021 by the xcube development team and contributors
```

xcube/cli/versions.py
```diff
@@ -0,0 +1,70 @@
...
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy of
```
TonioF, line 4: This line is a little bit too long. In xcube.util.versions I have adjusted it so it corresponds to our guidelines. We should use that whenever we create new modules.

xcube/cli/versions.py
```diff
@@ -0,0 +1,70 @@
...
+    if format_name.lower() == JSON_FORMAT:
+        import json
+        text = json.dumps(xcube_versions, indent=2)
+    else:
+        import yaml
+        text = yaml.dump(xcube_versions)
+
+    if output_path:
+        with open(output_path, 'w') as fp:
+            fp.write(text)
```
TonioF, line 68: I'd like a check whether the output path has the appropriate file extension. In case it hasn't at the very least  a warning could be given.

TonioF (commented):
See comments for details.

xcube/cli/versions.py
forman: > I'd like a check whether the output path has the appropriate file extension. In case it hasn't at the very least a warning could be given.

I disagree, because it is up to users if they use the `-o` option. Filename extensions are only conventions, not standards.

xcube/cli/versions.py
forman: Same for _all_ xcube modules. Let's fix this in one go later.

PR #535, opened 2021-09-21
Unify Dataset IO

TonioF:
This PR addresses #516. It ensures that, if a Configuration with Datasets is passed to a ServiceContext, each dataset is assigned to a datastore that will be used to open the data (unless the dataset is in-memory). It is attempted to find common roots for data stores to group datasets located at same/similar locations together.

When reviewing this, please consider especially the function get_data_store_pool of the ServerContext: This function used to remove a data store pool if there were no datastores explicitly configured. I had to change this, but I don't fully understand what was the original reasoning behind this.

This is a follow-up PR to erroneously merged PR #527 .

Checklist:

* [x] AppVeyor CI passes


forman (changes requested):
Where are the changes in `xcube.webapi.context`?

Also I don't understand your comment: 

> When reviewing this, please consider especially the function get_data_store_pool of the ServerContext: This function used to remove a data store pool if there were no datastores explicitly configured. I had to change this, but I don't fully understand what was the original reasoning behind this.

I cannot see what you are referring to.

forman (approved):

test/webapi/test_context.py
```diff
@@ -221,3 +221,246 @@ def test_interpolates_vars(self):
...
+                'Path': '/one/path/abc.zarr'
+            },
+            {
+                'Identifier': 'z_1',
+                'FileSystem': 'local',
+                'Path': '/one/path/def.zarr'
+            },
+            {
+                'Identifier': 'z_4',
+                'FileSystem': 'obs',
```
AliceBalfanz, line 243: I think we should stop using "obs" when referring to s3 object storage - and exchange it to "s3" as used in the DataStores configurations.

forman: Agreed!

TonioF: Correct - but for now, we'll move this to another PR.

Issue #538, opened 2021-10-04
JSON schema validation failing due to deprecation of "types" keyword


pont-us:
Since the release of jsonschema 4, multiple unit tests are failing in xcube, xcube-cci, and xcube-cds. The culprit is the `types` argument supplied to `jsonschema.validate` in `xcube.util.jsonschema.validate_instance`, which has been deprecated for a while and is now entirely unsupported as of jsonschema 4.

We use the `types` keyword to make jsonschema recognize tuples as arrays; this functionality now needs to be re-implemented in a jsonschema-4-compatible way.

pont-us:
Fixed by #539 .

PR #539, opened 2021-10-04
JSON schema validation: re-implement type checking

pont-us:
Addresses Issue #538.

JSON Schema 4 no longer supports the types keyword for validation.
This PR changes the implementation of schema validation to carry
out type checking with a custom validator instead.

[Description of PR]

Checklist:

* [ ] AppVeyor CI passes

### DIFF
--- a/xcube/util/jsonschema.py
+++ b/xcube/util/jsonschema.py
@@ -101,10 +101,26 @@ class JsonSchema(ABC):
         # jsconschema needs extra packages installed to validate some formats;
         # if they are missing, the format check will be skipped silently. For
         # date-time format, strict_rfc3339 or rfc3339-validator is required.
+
+        schema = self.to_dict()
+
+        # By default, jsonschema only recognizes lists as arrays. Here we derive
+        # and use a custom validator which recognizes both lists and tuples as
+        # arrays.
+        base_validator = jsonschema.validators.validator_for(schema)
+        new_type_checker = \
+            base_validator.TYPE_CHECKER.redefine(
+                'array',
+                lambda checker, inst: isinstance(inst, (list, tuple)))
+        custom_validator = \
+            jsonschema.validators.extend(base_validator,
+                                         type_checker=new_type_checker)
+
         jsonschema.validate(instance=instance, schema=self.to_dict(),
+                            cls=custom_validator,
                             format_checker=jsonschema.draft7_format_checker,
                             # We have to explicitly sanction tuples as arrays.
-                            types=dict(array=(list, tuple)))
+                            )
 
     def to_instance(self, value: Any) -> Any:
         """Convert Python object *value* into JSON value and return the validated result."""


xcube/util/jsonschema.py
```diff
@@ -101,10 +101,26 @@ def validate_instance(self, instance: Any):
...
+        base_validator = jsonschema.validators.validator_for(schema)
+        new_type_checker = \
+            base_validator.TYPE_CHECKER.redefine(
+                'array',
+                lambda checker, inst: isinstance(inst, (list, tuple)))
+        custom_validator = \
+            jsonschema.validators.extend(base_validator,
+                                         type_checker=new_type_checker)
+
         jsonschema.validate(instance=instance, schema=self.to_dict(),
```
forman, line 119:
```suggestion
        jsonschema.validate(instance=instance, schema=schema,
```

otherwise `schema.to_dict()` is computed twice

xcube/util/jsonschema.py
```diff
@@ -101,10 +101,26 @@ def validate_instance(self, instance: Any):
...
         # if they are missing, the format check will be skipped silently. For
         # date-time format, strict_rfc3339 or rfc3339-validator is required.
+
+        schema = self.to_dict()
+
+        # By default, jsonschema only recognizes lists as arrays. Here we derive
+        # and use a custom validator which recognizes both lists and tuples as
+        # arrays.
+        base_validator = jsonschema.validators.validator_for(schema)
+        new_type_checker = \
```
forman, line 111: This wiring up here is not clear to me. Is it really that complicated to add a custom type check?

xcube/util/jsonschema.py
```diff
@@ -101,10 +101,26 @@ def validate_instance(self, instance: Any):
...
+        custom_validator = \
+            jsonschema.validators.extend(base_validator,
+                                         type_checker=new_type_checker)
+
         jsonschema.validate(instance=instance, schema=self.to_dict(),
+                            cls=custom_validator,
                             format_checker=jsonschema.draft7_format_checker,
                             # We have to explicitly sanction tuples as arrays.
-                            types=dict(array=(list, tuple)))
+                            )
```
forman, line 123: Format. Avoid hanging last paren.

forman (changes requested):

xcube/util/jsonschema.py
TonioF: Assuming you mean self, this will lead to a lot of test failures.

xcube/util/jsonschema.py
pont-us: I agree, it's rather involved compared to the old method, but this really is officially The Right Way according to [the JSON Schema documentation](https://python-jsonschema.readthedocs.io/en/stable/validate/#validating-with-additional-types), and I couldn't find a simpler one.

xcube/util/jsonschema.py
pont-us: I'm probably missing something here, but I can't find the other invocation of `to_dict()`, `schema` isn't defined, and if `schema` is a typo for `self` I don't understand how jsonschema is supposed to validate a `JsonSchema` object rather than the dict it's expecting.

xcube/util/jsonschema.py
forman: Bah, I don't like that.

forman (approved):
Ok!

PR #566, opened 2021-11-17
handle non-existing roots

TonioF:
Avoids errors when a file system data store is created with a root that does not yet exist. The root will be created the first time that data is written to the store, it will be ignored beforehand.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,4 @@
-## Changes in 0.9.2
+## Changes in 0.9.2 (in development)
 
 ### Fixes
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+## Changes in 0.9.2
+
+### Fixes
+
+* Be more robust with respect to non-existing data store root directories.   
+
 ## Changes in 0.9.1
 
 ### New features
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -1,3 +1,9 @@
+## Changes in 0.9.2
```
AliceBalfanz, line 1:
```suggestion
## Changes in 0.9.2 (in development)
```

xcube/core/store/fs/store.py
```diff
@@ -581,6 +581,8 @@ def _generate_data_ids(self,
                            return_tuples: bool,
                            current_depth: int):
         root = self.root + ('/' + dir_path if dir_path else '')
+        if not self.fs.exists(root):
+            return
```
AliceBalfanz, line 585: Why return? where is the directory created in case it does not exist?

AliceBalfanz (changes requested):
Pls also set the version to 9.2.0 dev :) Let's first test and if it solves all our issues, then we can make a release

xcube/core/store/fs/store.py
TonioF: Return because no data ids exist

TonioF:
Yes, definitely let's first do a dev version.

PR #597, opened 2022-01-28
Fixed flipping for unevenly tiled images

TonioF:
Implemented changes after our discussion. Works nicely now!


forman (approved):
Fantastico!

PR #600, opened 2022-02-01
Replaced file system values obs and local with s3 and file

TonioF:
The file system identifiers 'obs' and 'local' have been replaced with the names of their respective protocols, 's3' and 'file'. The old values are still supported for backwards compatibility.


AliceBalfanz (approved):
Thank you for making the changes :)

forman (approved):

PR #602, opened 2022-02-02
Tile server: ensure timezone-naive time selector

### LINKED ISSUE #555: Address `xcube serve`  FutureWarning: timezone-naive DatetimeIndex with a timezone-aware datetime is deprecated
When serving xcube datasets with `xcube serve` there is a 
`FutureWarning: Indexing a timezone-naive DatetimeIndex with a timezone-aware datetime is deprecated and will raise KeyError in a future version.  Use a timezone-naive object instead.`
We need to address this before we run into that KeyError.

pont-us:
Fixes #555.

In tile._get_var_2d_array (used by new_color_mapped_image),
check for the possibility that the array uses datetime64 and the
indexer is timezone-aware. In this case, convert to a timezone-naive
indexer before performing the selection.

[Description of PR]

Checklist:

* ~~[ ] Add unit tests and/or doctests in docstrings~~ (not needed)
* ~~[ ] Add docstrings and API docs for any new/modified user-facing classes and functions~~ N/A
* ~~[ ] New/modified features documented in `docs/source/*`~~ N/A
* ~~[ ] Changes documented in `CHANGES.md`~~ N/A
* [x] AppVeyor CI passes

### DIFF
--- a/xcube/core/tile.py
+++ b/xcube/core/tile.py
@@ -242,8 +242,7 @@ def _get_var_2d_array(var: xr.DataArray,
         if labels_are_indices:
             array = var.isel(**labels)
         else:
-            labels_safe_time = _ensure_time_compatible(var, labels)
-            array = var.sel(method='nearest', **labels_safe_time)
+            array = var.sel(method='nearest', **labels)
     else:
         raise exception_type(f'Variable "{var.name}" of dataset "{ds_id}" '
                              'must be an N-D Dataset with N >= 2, '
@@ -328,7 +327,8 @@ def parse_non_spatial_labels(
         dims: Sequence[Hashable],
         coords: Mapping[Hashable, xr.DataArray],
         allow_slices: bool = False,
-        exception_type: type = ValueError
+        exception_type: type = ValueError,
+        var: xr.DataArray = None
 ) -> Mapping[str, Any]:
     xy_var_names = get_dataset_xy_var_names(coords, must_exist=False)
     if xy_var_names is None:
@@ -384,4 +384,7 @@ def parse_non_spatial_labels(
             raise exception_type(f'{label_str!r} is not a valid'
                                  f' value for dimension {dim!r}') from e
 
-    return parsed_labels
+    if var is not None:
+        return _ensure_time_compatible(var, parsed_labels)
+    else:
+        return parsed_labels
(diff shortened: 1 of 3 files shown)


AliceBalfanz (approved):
Tested locally and works well :) Thanks for fixing!

forman (changes requested):
- The natural home of this fix should be `xcube.core.tile.parse_non_spatial_labels()`
- I assume, the nature of the problem requires a more general solution. Hence, we need an understanding of the root cause and a general function that fixes the issue before timstamps are passed into `xarray.Dataset/DataArray.sel()`.

xcube/core/tile.py
```diff
@@ -242,14 +242,34 @@ def _get_var_2d_array(var: xr.DataArray,
...
+                            labels: Dict[str, Any]) -> Dict[str, Any]:
+    """Ensure that labels['time'] is timezone-naive, if necessary.
+    If var has a 'time' dimension of type datetime64 and labels has a 'time'
+    key with a timezone-aware value, return a modified labels dictionary with
+    a timezone-naive time value. Otherwise return the original labels.
+    """
+    if 'time' in var.dims and \
+        hasattr(var['time'], 'dtype') and \
+        hasattr(var['time'].dtype, 'type') and \
+        var['time'].dtype.type is np.datetime64 and \
```
TonioF, line 264: For readability, I'd prefer if this was moved into a helper function

TonioF (commented):
I really would like a test for this. Also, I'd expect this issue to arise in more places than this. Did you check the other places?

xcube/core/tile.py
pont-us: Seems sensible -- done.

pont-us:
> I really would like a test for this.

OK, added.

pont-us:
> Did you check the other places?

There seem to be fewer of them than I feared. I found time selection used in three other places in the xcube codebase:

 - `core.select.select_temporal_subset`
 - `core.timeseries.get_time_series`
 - `cli.tile.tile`

The last of these will be made safe by Norman's suggestion of moving the check to `core.tile.parse_non_spatial_labels()`, but the other two will require further work.

pont-us:
> The natural home of this fix should be `xcube.core.tile.parse_non_spatial_labels()`

OK, moved.

forman:
> but the other two will require further work.

Then let's just fix the server and enhance the issue #555 by your findings or create a new one.

PR #607, opened 2022-02-08
adapt to zarr 2.11

TonioF:
Fixes pruning tests

### DIFF
--- a/test/cli/test_prune.py
+++ b/test/cli/test_prune.py
@@ -1,7 +1,6 @@
 import os.path
 import sys
 
-import numpy as np
 import xarray as xr
 
 from test.cli.helpers import CliTest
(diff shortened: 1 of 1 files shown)


test/cli/test_prune.py
```diff
@@ -25,12 +24,24 @@ class PruneDataTest(CliTest):
     def setUp(self) -> None:
         rimraf(self.TEST_CUBE)
         cube = new_cube(time_periods=3,
-                        variables=dict(precipitation=np.nan,
-                                       temperature=np.nan)) \
+                        variables=dict(precipitation=1.0,
+                                       temperature=1.0)) \
```
forman, line 28: Why not continuing using `NaN` and setting `fill_value=null`?

forman (commented):
See comment.

forman (approved):

PR #614, opened 2022-02-22
Improved Error Handling

TonioF:
This PR is in reaction to repeated observations that xcube gen2 would return rather cryptic 500 errors. On cumbersome investigation, it often turned out that the error was not on server-side, but that the problem was a bad request which resulted in errors that were not handled (correctly).

Checklist:

* ~[ ] Add unit tests and/or doctests in docstrings~
* [x] AppVeyor CI passes

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,7 @@
 
 ### Enhancements
 
-* `xcube gen` returns more expressive error messages 
+* `xcube gen2` returns more expressive error messages 
 
 ### Fixes
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ### Enhancements
 
+* `xcube gen` returns more expressive error messages 
+
 ### Fixes
 
 ### Other
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,7 +8,7 @@
   value of `base_dataset_id`, instead of being copied as-is.
   This can save large amounts of storage space. (#617)
 
-* `xcube gen2` returns more expressive error messages
+* `xcube gen2` returns more expressive error messages.
   
 ### Fixes
 
--- a/xcube/core/gen2/local/describer.py
+++ b/xcube/core/gen2/local/describer.py
@@ -73,7 +73,7 @@ class DatasetsDescriber:
                                               data_type=DATASET_TYPE)
         except DataStoreError as dse:
             raise CubeGeneratorError(f'{dse}',
-                                     status_code=400)
+                                     status_code=400) from dse
         if not isinstance(descriptor, DatasetDescriptor):
             raise RuntimeError(f'internal error: data store '
                                f'"{input_config.store_id}": '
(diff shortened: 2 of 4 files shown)


xcube/core/gen2/local/opener.py
```diff
@@ -59,41 +62,47 @@ def open_cube(self, input_config: InputConfig) -> TransformedCube:
...
+                dataset = opener.open_data(input_config.data_id,
+                                           **open_params,
+                                           **dataset_open_params)
+                observer.worked(1)
+
+            except DataStoreError as dse:
+                tb = traceback.format_tb(dse.__traceback__)
+                raise CubeGeneratorError(f'{dse}',
+                                         status_code=400,
+                                         remote_traceback=tb)
```
forman, line 105: `remote_traceback` is not appropriate in package `local`

xcube/core/gen2/local/opener.py
```diff
@@ -59,41 +62,47 @@ def open_cube(self, input_config: InputConfig) -> TransformedCube:
...
+
+                open_params_schema = opener.get_open_data_params_schema(
+                    input_config.data_id
+                )
 
-            observer.worked(1)
+                dataset_open_params = {
+                    k: v for k, v in cube_params.items()
+                    if k in open_params_schema.properties
+                }
```
forman, line 92: I do not understand any of the changes above. What's the reason for this?

TonioF: This is all about catching DataStoreErrors. As there are several instances where one could be raised, I wrapped several lines of a code in a try block. You could think about moving these lines to a dedicated method, however, I did not want to do this as it is also wrapped in a with-block.

xcube/core/gen2/local/describer.py
```diff
@@ -71,10 +71,8 @@ def _describe_dataset(self, input_config: InputConfig) \
         try:
             descriptor = opener.describe_data(input_config.data_id,
                                               data_type=DATASET_TYPE)
-        except DataStoreError:
-            raise CubeGeneratorError(f'Data store '
-                                     f'"{input_config.store_id}" '
-                                     f'does not support datasets',
+        except DataStoreError as dse:
+            raise CubeGeneratorError(f'{dse}',
                                      status_code=400)
```
forman, line 76: Keep original traceback:

```suggestion
            raise CubeGeneratorError(f'{dse}',
                                     status_code=400) from dse
```

CHANGES.md
```diff
@@ -8,6 +8,8 @@
   value of `base_dataset_id`, instead of being copied as-is.
   This can save large amounts of storage space. (#617)
 
+* `xcube gen2` returns more expressive error messages
```
forman, line 11:
```suggestion
* `xcube gen2` returns more expressive error messages.
```

forman (changes requested):

AliceBalfanz (approved):

Issue #615, opened 2022-02-22
Support different resolutions in x- and y- direction


TonioF:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #621, opened 2022-02-24
support different resolutions in x- and y- direction

TonioF:
Solves #615 . DatasetDescriptors may hold a tuple of two values for spatial resolution, the resolution determined from a dataset may consist of two values, different resolutions are considered during cube generation.
Checklist:

* [x] AppVeyor CI passes

### DIFF
--- a/xcube/core/gen2/local/informant.py
+++ b/xcube/core/gen2/local/informant.py
@@ -31,7 +31,7 @@ from xcube.core.store import DataStorePool
 from xcube.core.store import DatasetDescriptor
 from xcube.core.store import VariableDescriptor
 from xcube.util.assertions import assert_instance
-from xcube.util.assertions import assert_true
+from xcube.util.types import normalize_number_scalar_or_pair
 from .describer import DatasetsDescriber
 from ..config import CubeConfig
 from ..request import CubeGeneratorRequest
@@ -211,17 +211,9 @@ class CubeInformant:
         spatial_res = cube_config.spatial_res
         if spatial_res is None:
             spatial_res = self.first_input_dataset_descriptor.spatial_res
-        assert_instance(spatial_res, (numbers.Number, tuple), 'spatial_res')
-        if isinstance(spatial_res, tuple):
-            assert_true(len(spatial_res) == 2,
-                        'spatial_res must hold exactly two values when given '
-                        'as tuple')
-            assert_instance(spatial_res[0], numbers.Number, 'spatial_res')
-            assert_instance(spatial_res[1], numbers.Number, 'spatial_res')
-            assert_true(spatial_res[0] > 0, 'spatial_res must be positive')
-            assert_true(spatial_res[1] > 0, 'spatial_res must be positive')
-        else:
-            assert_true(spatial_res > 0, 'spatial_res must be positive')
+        spatial_res = normalize_number_scalar_or_pair(spatial_res, float)
+        assert_instance(spatial_res[0], numbers.Number, 'spatial_res')
+        assert_instance(spatial_res[1], numbers.Number, 'spatial_res')
 
         tile_size = cube_config.tile_size
         if tile_size is not None:
--- a/xcube/core/gen2/local/informant.py
+++ b/xcube/core/gen2/local/informant.py
@@ -209,8 +209,8 @@ class CubeInformant:
         if spatial_res is None:
             spatial_res = self.first_input_dataset_descriptor.spatial_res
         spatial_res = normalize_scalar_or_pair(spatial_res,
-                                               (int, float),
-                                               'spatial_res')
+                                               item_type=(int, float),
+                                               name='spatial_res')
         assert_instance(spatial_res[0], numbers.Number, 'spatial_res')
         assert_instance(spatial_res[1], numbers.Number, 'spatial_res')
 
--- a/xcube/core/gen2/local/informant.py
+++ b/xcube/core/gen2/local/informant.py
@@ -80,8 +80,12 @@ class CubeInformant:
         x_min, y_min, x_max, y_max = cube_config.bbox
         spatial_res = cube_config.spatial_res
 
-        width = round((x_max - x_min) / spatial_res)
-        height = round((y_max - y_min) / spatial_res)
+        if isinstance(spatial_res, tuple):
+            width = round((x_max - x_min) / spatial_res[0])
+            height = round((y_max - y_min) / spatial_res[1])
+        else:
+            width = round((x_max - x_min) / spatial_res)
+            height = round((y_max - y_min) / spatial_res)
         width = 2 if width < 2 else width
         height = 2 if height < 2 else height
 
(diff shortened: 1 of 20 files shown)


forman (changes requested):
- Don't repeat yourself, use normalizer function where applicable
- Be pythonic, expect a scalar of type T or a pair of T, don't check for instanceof `tuple`

Type annotation `Tuple[T, T]` should be understooof as `Pair[T]`. Suggestion: let's add to xcube 
  - type annotations for both `Pair[T]` and `ScalarOrPair[T]`;
  - function `normalize_number_scalar_or_pair(sop: ScalarOrPair[T], t: Type[T]) -> Pair[T]`.

forman (approved):

xcube/core/gen2/local/informant.py
```diff
@@ -207,8 +211,9 @@ def _compute_effective_cube_config(self) \
         spatial_res = cube_config.spatial_res
         if spatial_res is None:
             spatial_res = self.first_input_dataset_descriptor.spatial_res
-        assert_instance(spatial_res, numbers.Number, 'spatial_res')
-        assert_true(spatial_res > 0, 'spatial_res must be positive')
+        spatial_res = normalize_number_scalar_or_pair(spatial_res, float)
```
forman, line 214:
```suggestion
        spatial_res = normalize_number_scalar_or_pair(spatial_res)
```

forman (changes requested):
Much better now. However, I would have expected a generic solution for the pair normalisation as discussed, for example:

```python
def normalize_scalar_or_pair(
    value: ScalarOrPair[T], 
    item_type: Optional[Type[T]] = None, 
    name: Optional[str] = None
):
    try:
        x, y = value
    except TypeError:
        x, y = value, value
    if item_type is not None:
        assert_true(isinstance(x, item_type) and isinstance(y, item_type),
                           message=f"{name or 'value'} must be a scalar"
                            " or pair of {item_type}")
    return x, y
```

Then

```
spatial_res = normalize_scalar_or_pair(spatial_res, item_type=(int, float), name="spatial_res")
```

but also

```
tile_size = normalize_scalar_or_pair(tile_size, item_type=int, name="tile_size")
```

xcube/util/types.py
```diff
@@ -0,0 +1,43 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+from typing import Tuple, Type, TypeVar, Union
+
+T = TypeVar('T', int, float)
+
+Pair = Tuple[T, T]
+ScalarOrPair = Union[T, Pair]
+Bbox = Tuple[T, T, T, T]
```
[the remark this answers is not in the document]
TonioF: You're right, I'm removing this.

xcube/util/types.py
```diff
@@ -0,0 +1,75 @@
...
+
+T = TypeVar('T', int, float)
+
+ItemType = Union[Type[T], Tuple[Type[T], ...]]
+Pair = Tuple[T, T]
+ScalarOrPair = Union[T, Pair]
+
+
+def normalize_scalar_or_pair(
+        value: Optional[ScalarOrPair] = None,
```
forman, line 34: Value must not be optional and therefore it should be the first positional argument. 

```suggestion
        value: ScalarOrPair[T],
```

xcube/util/types.py
```diff
@@ -0,0 +1,75 @@
...
+    try:
+        assert_true(len(value) <= 2,
+                    message=f"{name or 'Value'} must be a scalar or pair of "
+                            f"{item_type or 'scalars'}, was '{value}'")
+        x, y = value
+    except TypeError:
+        x, y = value, value
+    if item_type is not None:
+        x = _maybe_convert(x, item_type)
+        y = _maybe_convert(y, item_type)
```
forman, line 53: 1. I don't know why type conversion should take place at all. Where does the requirement come from?
2. If `item_type` is a tuple, the converted result is dependent on the order of items in the type tuple. Is this intended? If so, why?

xcube/util/types.py
```diff
@@ -0,0 +1,75 @@
...
+ItemType = Union[Type[T], Tuple[Type[T], ...]]
+Pair = Tuple[T, T]
+ScalarOrPair = Union[T, Pair]
+
+
+def normalize_scalar_or_pair(
+        value: Optional[ScalarOrPair] = None,
+        *,
+        item_type: Optional[ItemType[T]] = None,
+        default: Optional[ScalarOrPair] = None,
```
forman, line 37: To me, a `default` value does not make sense, because the function is about _normalization_ of `value`.
In addition, it is fine for `value` to be `None` as a value, not as an indicator for absense. The result would be `(None, None)` which may be desired.

xcube/util/types.py
```diff
@@ -0,0 +1,75 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+from typing import Optional, Tuple, Type, TypeVar, Union
+
+from xcube.util.assertions import assert_true
+
+T = TypeVar('T', int, float)
```
forman, line 26: T shall not be limited to `int` and `float`, but any Python type:

```suggestion
T = TypeVar('T')
```

forman (changes requested):
All fine, but new module `types`. I think, still lacks generality and makes some incomprehensible assumtions. See my comments.

xcube/util/types.py
TonioF: > 1. I don't know why type conversion should take place at all. Where does the requirement come from?

This is what I would expect from a normalization. If you hand in one or more types you want the resulting data to be of one of these types. Otherwise it would be a validation. 
This is also in line with other normalization methods, e.g., https://github.com/dcs4cop/xcube/blob/d3fc869ce1a80541d876cb1a0b484b2bb5956dd0/xcube/core/gridmapping/helpers.py#L89

> 2. If `item_type` is a tuple, the converted result is dependent on the order of items in the type tuple. Is this intended? If so, why?

I am aware of this, but I don't see a reason why this should be done differently. Adding documentation would be a good idea, though.

xcube/util/types.py
TonioF: > To me, a `default` value does not make sense, because the function is about _normalization_ of `value`. 

Again, this is in line with existing normalization behaviour (same link as above): https://github.com/dcs4cop/xcube/blob/d3fc869ce1a80541d876cb1a0b484b2bb5956dd0/xcube/core/gridmapping/helpers.py#L89

There would be uses for a helper function that applies a default if the normalization value is not valid. Maybe this could be a second function?

> In addition, it is fine for `value` to be `None` as a value, not as an indicator for absense. The result would be `(None, None)` which may be desired.

Agreed.

TonioF:
I want to keep the type checking and you bring up a good argument against type conversion.

With respect to defaults: I'd like to keep it in, too, but I am fine with removing it here and reinstantiating the helper functions.

forman (approved):
Yeah!

PR #639, opened 2022-03-08
Allow running xcube Env with Coiled

forman:
Added lz4 and python-blosc (should be transitive deps anyway), so we can use xcube env on Coiled without modifications

Checklist:

* [x] AppVeyor CI passes

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -49,10 +49,14 @@
 
 ### Other
 
-* Replace the dependency on the rfc3339-validator PyPI package with a
+* Added packages `python-blosc` and `lz4` to the xcube Python environment 
+  for better support of Dask `distributed` and the Dask service 
+  [Coiled](https://coiled.io/).
+
+* Replace the dependency on the `rfc3339-validator` PyPI package with a
   dependency on its recently created conda-forge package.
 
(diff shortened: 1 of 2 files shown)


forman:
FYI @AliceBalfanz @pont-us

TonioF (changes requested):
Both packages were not included in my local xcube environment, hence I do not think they are "transitive deps anyway".
What troubles me most though are the many test failures on macos according to appveyor. If these failures are related to these changes (which I strongly assume), I suggest that you create an alternative environment.yml somewhere else and use that one for coiled.

forman:
> If these failures are related to these changes (which I strongly assume)

They are not. xcube main branch has the same issues. @pont-us already reported them yesterday.

TonioF:
> > If these failures are related to these changes (which I strongly assume)
> 
> They are not. xcube main branch has the same issues. @pont-us already reported them yesterday.

I checked whether the errors also appear on the main branch, which they did not, as they are already fixed there. Can you please merge the main branch so we have a complete appveyor check?

forman:
I'll close this as this already way to much effort. Will create a dedicated env for time being.

forman:
Need to reopen, as there is one already dependent on this.

TonioF (approved):

PR #647, opened 2022-03-15
Pyramids may contain any variables

### LINKED ISSUE #646: Level 0 of generated pyramids is lacking spatial CRS
**Describe the bug**

The level datasets of generated dataset pyramids only comprise spatial variables. All other variables are filtered out including variables that represent the spatial variables' spatial CRS. At least level 0 should contain also non-spatial variables

**To Reproduce**
```
ds = data_store.open_data("test.zarr")
data_store.write_data(ds, "test.levels")
ml_ds = data_store.open_data("test.levels") --> ERROR
```

The multi-level dataset cannot be opened because it is lacking a CRS.

forman:
In this PR:

* Fixed problem where the dataset levels of multi-level datasets were written without spatial coordinate reference system. In fact, only spatial variables were written.

* Added new module `xcube.core.subsampling` for function `subsample_dataset(dataset, step)` that is now used by default to generate the datasets level of multi-level datasets.

Closes #646.

Checklist:

* [ ] AppVeyor CI passes

### DIFF
--- a/xcube/core/subsampling.py
+++ b/xcube/core/subsampling.py
@@ -91,7 +91,6 @@ def get_dataset_subsampling_slices(
     for var_name, var in dataset.data_vars.items():
         var_index = slices_dict.get(var.dims)
         if var_index is None:
-            var_index = None
             for index, dim_name in enumerate(var.dims):
                 if dim_name == x_dim_name or dim_name == y_dim_name:
                     if var_index is None:
--- a/xcube/core/subsampling.py
+++ b/xcube/core/subsampling.py
@@ -1,0 +1,107 @@
+#  The MIT License (MIT)
+#  Copyright (c) 2022 by the xcube development team and contributors
+#
+#  Permission is hereby granted, free of charge, to any person obtaining a
+#  copy of this software and associated documentation files (the "Software"),
+#  to deal in the Software without restriction, including without limitation
+#  the rights to use, copy, modify, merge, publish, distribute, sublicense,
+#  and/or sell copies of the Software, and to permit persons to whom the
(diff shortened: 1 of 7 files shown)


xcube/core/subsampling.py
```diff
@@ -0,0 +1,107 @@
...
+    """
+    assert_instance(dataset, xr.Dataset, name='dataset')
+    assert_instance(step, int, name='step')
+    x_dim_name, y_dim_name = xy_dim_names or ('x', 'y')
+    slices_dict: Dict[Tuple[Hashable, ...], Tuple[slice, ...]] = dict()
+    vars_dict: Dict[Hashable, Optional[Tuple[slice, ...]]] = dict()
+    for var_name, var in dataset.data_vars.items():
+        var_index = slices_dict.get(var.dims)
+        if var_index is None:
+            var_index = None
```
TonioF, line 94:
```suggestion
```

xcube/core/subsampling.py
```diff
@@ -0,0 +1,107 @@
...
+        data_vars={
+            var_name: (var[var_slices[var_name]]
+                       if var_name in var_slices else var)
+            for var_name, var in dataset.data_vars.items()
+        },
+        attrs=dataset.attrs
+    )
+
+
+_EMPTY_SLICE = slice(None, None, None)
```
TonioF, line 69: I see that this name makes sense, but for understanding what is the purpose of this slice, the term _FULL_SLICE would be better. I'd definitely prefer it.

TonioF (approved):
Works well. I have two comments, though, neither of which is crucial.

Issue #651, opened 2022-03-17
xcube docker image is not working


sfalquier:
**Describe the bug**
Following the [documentation](https://xcube.readthedocs.io/en/latest/installation.html#docker), after building xcube docker image from repository (v0.10.2), we are not able to run directly xcube commands in container and bound port does not correspond with image's exposed port.

**To Reproduce**
```
$ docker build -t xcube:latest .
$ docker run -d -p 8001:8000 xcube:latest xcube
8635633f3b9b122d2ee3836cf4afe91af2f486933c4403c22a041475d3637495
docker: Error response from daemon: failed to create shim: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: "xcube": executable file not found in $PATH: unknown.
```
It looks like `source activate xcube` is not taken into account when command is executed at runtime in the container.

**Expected behavior**
Something like this with correct port mapping : 
```
$ docker build -t xcube:latest .
$ docker run -d -p **8080:8080** xcube:latest xcube
8635633f3b9b122d2ee3836cf4afe91af2f486933c4403c22a041475d3637495
Usage: xcube [OPTIONS] COMMAND [ARGS]...

  xcube Toolkit

Options:
  --version              Show the version and exit.
  --traceback            Enable tracing back errors by dumping the Python call
                         stack. Pass as very first option to also trace back
                         error during command-line validation.
  --scheduler SCHEDULER  Enable distributed computing using the Dask scheduler
                         identified by SCHEDULER. SCHEDULER can have the form
                         <address>?<keyword>=<value>,... where <address> is
                         <host> or <host>:<port> and specifies the scheduler's
                         address in your network. For more information on
                         distributed computing using Dask, refer to
                         http://distributed.dask.org/. Pairs of
                         <keyword>=<value> are passed to the Dask client.
                         Refer to http://distributed.dask.org/en/latest/api.ht
                         ml#distributed.Client
  -w, --warnings         Show any warnings produced during operation (warnings
                         are hidden by default)
  --help                 Show this message and exit.

Commands:
  chunk     (Re-)chunk xcube dataset.
  compute   Compute a cube from one or more other cubes.
  dump      Dump contents of an input dataset.
  edit      Edit the metadata of an xcube dataset.
  extract   Extract cube points.
  gen       Generate xcube dataset.
  grid      Find spatial xcube dataset resolutions and adjust bounding...
  level     Generate multi-resolution levels.
  optimize  Optimize xcube dataset for faster access.
[...]

PR #664, opened 2022-04-07
Docker image is not working

TonioF:
Adresses #651. I reintroduced an entrypoint which is running in the xcube conda enviroment, the demo service is started by default. Also, I updated the documentation.

Checklist:

* ~[ ] Add unit tests and/or doctests in docstrings+
* [x] AppVeyor CI passes


Dockerfile
```diff
@@ -58,5 +58,8 @@ RUN if [[ ${INSTALL_PLUGINS} == '1' ]]; then bash install_xcube.sh xcube-cds ${X
 # Export web server port
 EXPOSE 8080
 
-# Start bash, so we can invoke xcube CLI.
-CMD ["/bin/bash"]
+# Run bash in xcube environment, so we can invoke xcube CLI.
+ENTRYPOINT ["conda", "run", "-v", "-n", "xcube", "/bin/bash", "-c"]
+
+# By default run demo service.
+CMD ["xcube serve -v --address 0.0.0.0 --port 8080 -c /home/xcube/examples/serve/demo/config.yml"]
```
forman, line 65: Why running the demo service and not just bash with xcube env activated?

forman (changes requested):

Dockerfile
TonioF: This is an arbitrary choice. I chose this because it used to be the default. We could do either, or we could run "xcube --help" for starters. 
So, you prefer not to have a default command?

dzelge (commented):
See comments for ENTRYPOINT. I woudl be happy with the current version just wanted to give an alternative.

Dockerfile
```diff
@@ -58,5 +58,8 @@ RUN if [[ ${INSTALL_PLUGINS} == '1' ]]; then bash install_xcube.sh xcube-cds ${X
 # Export web server port
 EXPOSE 8080
 
-# Start bash, so we can invoke xcube CLI.
-CMD ["/bin/bash"]
+# Run bash in xcube environment, so we can invoke xcube CLI.
+ENTRYPOINT ["conda", "run", "-v", "-n", "xcube", "/bin/bash", "-c"]
+
+# By default run demo service.
```
forman, line 64:
```suggestion
# By default show xcube help 
```

forman (changes requested):
Just that tiny typo.

Dockerfile
forman: > My suggestion would be to use only (no COMMAND):

This way, we can't use the `xcube` specific options like `--traceback`, right?

PR #679, opened 2022-05-05
Enhance xcube affine_transform_dataset()

forman:
`xcube.core.resampling.affine_transform_dataset()` has a new keyword argument `reuse_coords: bool = False`. If set to `True` the returned dataset will reuse the _same_ spatial coordinates as the target. This is a workaround for xarray issue https://github.com/pydata/xarray/issues/6573.

Checklist:

* [ ] AppVeyor CI passes

### DIFF
--- a/test/core/gridmapping/test_coords.py
+++ b/test/core/gridmapping/test_coords.py
@@ -10,14 +10,17 @@ from xcube.core.gridmapping import GridMapping
 # noinspection PyProtectedMember
 
 GEO_CRS = pyproj.crs.CRS(4326)
+NOT_A_GEO_CRS = pyproj.crs.CRS(5243)
 
 
 # noinspection PyMethodMayBeStatic
 class Coords1DGridMappingTest(unittest.TestCase):
 
     def test_1d_j_axis_down(self):
-        gm = GridMapping.from_coords(x_coords=xr.DataArray(np.linspace(1.5, 8.5, 8), dims='lon'),
-                                     y_coords=xr.DataArray(np.linspace(4.5, -4.5, 10), dims='lat'),
+        x_coords = xr.DataArray(np.linspace(1.5, 8.5, 8), dims='lon')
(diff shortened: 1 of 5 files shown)


forman:
FYI @TonioF (get well soon!)

test/core/gridmapping/test_coords.py
```diff
@@ -121,21 +130,39 @@ def test_1d_xy_coords(self):
         self.assertEqual(('lon', 'lat'), gm.xy_var_names)
         self.assertEqual(('lon', 'lat'), gm.xy_dim_names)
 
+    def test_to_coords(self):
```
martin-boettcher, line 133: This is the new test for the parameter reuse_coords=True. Is it sufficient to test for the dtype of the result here because a newly generated set of coordinates would have been float64? If yes, would it have been a solution that the newly created dimensions have dtype float32 as well instead of float64, e.g. by preserving the dtype of the input coordinates? - For some reason the default of reuse_coords is False, so there will be cases where you want to not reuse them.

xcube/core/gridmapping/coords.py
```diff
@@ -311,6 +324,27 @@ def grid_mapping_to_coords(
...
+                and y.size == grid_mapping.height:
+            return {
+                name: xr.DataArray(coord.values,
+                                   dims=dim,
+                                   attrs=coord.attrs)
+                for name, dim, coord in zip(xy_var_names,
+                                            xy_dim_names,
+                                            (x, y))
+            }
+
```
martin-boettcher, line 347: The line I am referring to in my other comment is the
dtype = np.float64
below. Could this instead be derived from grid_mapping.x_coords' type?

martin-boettcher (approved):

xcube/core/gridmapping/coords.py
forman: I'll fix this in another PR later.

Issue #678, opened 2022-05-03
xcube serve does not provide datasets after service configuration changes


TonioF:
**Describe the bug**
When you have a running service, you may edit the service configuration. When you do this, any incoming requests to the service will no longer detect any datasets. The service will state that it does not provide any.

**To Reproduce**
1. Start an xcube service, pass a service configuration file as a parameter
2. Start an xcube viewer instance listening to the service
3. Edit the service configuration file, make sure to leave in valid data store configurations
4. Wait a short time until the new config has been ingested into the service (this event will be logged)
5. Reload the viewer.
No more datasets are shown, even though they are defined in the config.

**Expected behavior**
We would expect the service to provide the datasets as configured in the configuration file, in particular it should reflect the changes of the configuration file

TonioF:
Solved by #683

PR #683, opened 2022-05-10
xcube serve does not provide datasets after service configuration changes

TonioF:
Solves #678. The changes in this PR ensure that there is always a DataStorePool in the service context, and that the dataset_configs correspond to its content. The DataStorePool is repopulated when it has been emptied, which usually happens after an external config change.

Checklist:

* [ ] AppVeyor CI passes

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,8 +23,8 @@
 * Fixed ESA CCI example notebook. (#680)
 
 * `xcube serve` now provides datasets after changes of the service 
-  configuration.
-  Previously, it was necessary to restart a service to load the changes. (#678)
+  configuration while the server is running.
+  Previously, it was necessary to restart the server to load the changes. (#678)
 
 ### Other changes
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,6 @@
-### Enhancements
-
 ## Changes in 0.11.2 (in development)
+
+### Enhancements
 
 * `xcube serve` now publishes the chunk size of a variable's 
   time dimension for either for an associated time-chunked dataset or the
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -22,6 +22,10 @@
 
 * Fixed ESA CCI example notebook. (#680)
 
+* `xcube serve` now provides datasets after changes of the service 
+  configuration.
```
forman, line 26:
```suggestion
  configuration while the server is running.
```

CHANGES.md
```diff
@@ -22,6 +22,10 @@
...
+  Previously, it was necessary to restart a service to load the changes. (#678)
```
forman, line 27:
```suggestion
  Previously, it was necessary to restart the server to load the changes. (#678)
```

xcube/webapi/context.py
```diff
@@ -303,16 +302,12 @@ def get_variable_for_z(self, ds_id: str, var_name: str, z_index: int) -> xr.Data
...
-        if self._dataset_configs is None:
-            with self._lock:
-                dataset_configs = self._config.get('Datasets', [])
-                dataset_configs += \
-                    self.get_dataset_configs_from_stores()
-                self._dataset_configs = dataset_configs
-                self._maybe_assign_store_instance_ids()
+        with self._lock:
+            if self._dataset_configs is None:
+                self._set_up_data_store_pool_and_dataset_configs()
```
forman, line 307: Only lock, if needed. Double-check to be on the safe side (see https://en.wikipedia.org/wiki/Double-checked_locking).

```suggestion
        if self._dataset_configs is None:
            with self._lock:
                if self._dataset_configs is None:
                    self._set_up_data_store_pool_and_dataset_configs()
```

xcube/webapi/context.py
```diff
@@ -499,24 +486,34 @@ def new_dataset_metadata(self,
         # noinspection PyTypeChecker
         return dataset_metadata
 
-    def get_data_store_pool(self) -> Optional[DataStorePool]:
+    def get_data_store_pool(self) -> DataStorePool:
+        with self._lock:
```
forman, line 490: See above, use double-checked locking.

forman (changes requested):
Good! But please update according to my suggestions.

AliceBalfanz (approved):
From a pure usability aspect, this now works well. Tested locally.

PR #685, opened 2022-05-13
Webapi to provide dataset geometry

forman:
xcube Server now provides new metadata details of a dataset:
- The spatial reference is now given by property `spatialRef` and provides a textual representation of the spatial CRS.
- The dataset boundary is now given as property `geometry` and provides a GeoJSON Polygon in geographic coordinates. 

Checklist:

* [x] AppVeyor CI passes

Required by https://github.com/dcs4cop/xcube-viewer/pull/232

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+## Changes in 0.11.3 (in development)
+
+* xcube Server now provides a shorthand for the spatial reference 
+  in its dataset detail responses. The property `spatial_ref` now holds a
+  textual representation of the spatial CRS.
+
 ## Changes in 0.11.2
 
 ### Enhancements
@@ -38,8 +44,7 @@
   - `is_dataset_y_axis_inverted()` is no longer used;
   - `get_geometry_mask()` is no longer used;
   - `convert_geometry()` has been renamed to `normalize_geometry()`.
-
-
+  
 ## Changes in 0.11.1
 
 * Fixed broken generation of composite RGBA tiles. (#668)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,15 +1,13 @@
-## Changes in 0.11.3 (in development)
+## Changes in 0.11.2 (in development)
 
-* xcube Server now provides new metadata details of a dataset:
+### Enhancements
+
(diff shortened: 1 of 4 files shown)


AliceBalfanz (approved):
Works well, thank you for adding this enhancement!

xcube/webapi/controllers/catalogue.py
```diff
@@ -45,6 +44,7 @@
 from xcube.webapi.controllers.tiles import get_tile_source_options
 from xcube.webapi.errors import ServiceBadRequestError
 
+_CRS84 = pyproj.CRS.from_string('CRS84')
```
TonioF, line 48:
```suggestion


```

Just add empty line

CHANGES.md
```diff
@@ -1,3 +1,11 @@
+## Changes in 0.11.3 (in development)
+
+* xcube Server now provides new metadata details of a dataset:
+  - The spatial reference is now given by property `spatialRef` 
+    and provides a textual representation of the spatial CRS.
+  - The dataset boundary is now given as property `geometry`
+    and provides a GeoJSON Polygon in geographic coordinates. 
+
 ## Changes in 0.11.2
```
TonioF, line 9: Adapt. We'll make it to 0.11.2

xcube/webapi/controllers/catalogue.py
```diff
@@ -316,6 +323,28 @@ def get_dataset(ctx: ServiceContext,
     return dataset_dict
 
 
+def get_bbox_geometry(dataset_bounds: Tuple[float, float, float, float],
```
TonioF, line 326: I could well imagine this method as a utility function someplace else. Also, a dedicated unit test would have been nice.

TonioF (changes requested):
Two very minor issues and a remark.

xcube/webapi/controllers/catalogue.py
forman: Done.

forman: Right, once it is used a second time, we'll do so. Currently it is called unconditionally, therefore test is not required w.r.t. coverage.

PR #691, opened 2022-05-24
Fix broken env for Mac

forman:
Fixed the xcube environment for all platforms (Mac build was broken due to libgdal shared lib):

* Removed all upper bounds of package versions.
* Using latest version of mamba 
* Fixed new test failures due to `pyproj` upgrade


forman:
FYI @AliceBalfanz @TonioF  @TejasMorbagal

pont-us (approved):
I'm glad to finally see those version pins removed :).

A few tests are failing in my local (combined xcube + stores + other packages) environment, so there may be versioning interactions we will have to deal with outside of the xcube core package. But since everything's passing in a "pure" xcube environment, I see no reason not to approve.

PR #677, opened 2022-04-29
Server redesign

forman:
### Enhancements
* xcube Server has been rewritten almost from scratch.
  
  - Introduced a new endpoint `${server_url}/s3/{bucket}` that emulates
    and AWS S3 object storage for the published datasets.
    The `bucket` name can be either:
    * `datasets` - publishes all datasets in Zarr format.
    * `pyramids` - publishes all datasets in a multi-level `levels`
      format (multi-resolution N-D images)
      that comprises level datasets in Zarr format.
    
    With the new S3 endpoints in place, xcube Server instances can be used
    as xcube data stores as follows:
    
    ```python
    store = new_data_store(
        "s3", 
        root="datasets",   # bucket "datasets", use also "pyramids"
        max_depth=2,       # optional, but we may have nested datasets
        storage_options=dict(
            anon=True,
            client_kwargs=dict(
                endpoint_url='http://localhost:8080/s3' 
            )
        )
    )
    ```

  - The limited `s3bucket` endpoints are no longer available and are 
    replaced by `s3` endpoints. 

* xcube Server's colormap management has been improved in several ways:
  - Colormaps are no longer managed globally. E.g., on server configuration 
    change, new custom colormaps are reloaded from files. 
  - Colormaps are loaded dynamically from underlying 
    matplotlib and cmocean registries, and custom SNAP color palette files. 
    That means, latest matplotlib colormaps are now always available. (#687)
  - Colormaps can now be reversed (name prefix `"_r"`), 
    can have alpha blending (name prefix `"_alpha"`),
    or both (name prefix `"_r_alpha"`).
  - Loading of custom colormaps from SNAP `*.cpd` has been rewritten.
    Now also the `isLogScaled` property of the colormap is recognized. (#661)
  - The module `xcube.util.cmaps` has been redesigned and now offers
    three new classes for colormap management:
    * `Colormap` - a colormap 
    * `ColormapCategory` - represents a colormap category
    * `ColormapRegistry` - manages colormaps and their categories
  
### Other

* Deprecated CLI `xcube tile` has been removed.
* Deprecated modules and their references have finally been removed:
  - `xcube.core.tilegrid`
  - `xcube.core.tiledimage`
* The following functions have been deprecated:
  - `xcube.util.cmaps.get_cmap()`
  - `xcube.util.cmaps.get_cmaps()`

Closes #661
Closes #676
Closes #687

Checklist:

* [x] Add unit tests and/or doctests in docstrings
[...]

PR #693, opened 2022-05-27
Avoid using timezone-aware indexers on timezone-naive arrays

### LINKED ISSUE #605: Avoid using timezone-aware indexers on timezone-naive arrays
# The fundamental problem

Recent pandas versions (>=1.4.0, I believe) do not support indexing a timezone-naive array with a timezone-aware indexer, or vice versa. The same restriction applies to xarray's `DataArray.sel` method, which uses pandas. There are several places in the xcube codebase where such an incompatible ~sel~ operation might occur, and one place where we have already encountered and fixed a bug due to such an operation (Issue #555).

# The `xcube serve` bug (Issue #555)

## Description

In this issue, the xcube API serve was serving a dataset with a time dimension of type `datetime64`, and received a tile request for a timestamp specified as the string `2020-08-10T10:36:29Z`. Result:

```
...lib/python3.9/site-packages/pandas/core/indexes/base.py", line 5894, in _get_indexer_non_comparable
TypeError: Cannot compare dtypes datetime64[ns] and datetime64[ns, UTC]
```

## Analysis

In this case, the root cause within our own code is in `xcube.webapi.controllers.catalogue.get_dataset`, which calls `xcube.webapi.controllers.catalogue.get_dataset_coordinates`, which, for time co-ordinates, calls `xcube.core.timecoord.timestamp_to_iso_string`, which explicitly appends a `Z` to its output, making it a timezone-aware UTC time specification even if the dtype of the time variable in the dataset is timezone-naive. Ideally, if this were fixed, the xcube viewer would read the timezone-naive co-ordinates via the xcube-server API (`datasets/<dataset_name>` endpoint, `dimensions.coordinates` path in the JSON) and respect this timezone-naiveness in its time parameter when requesting the tile. However, in practice, timezone-awareness is evidently hardcoded into viewer tile requests: `getTileLayer` in `src/selectors/controlSelectors.tsx` constructs the time string using [Date.toISOString](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString), which always produces a timezone-aware UTC string.

## Minimal example

You can reproduce the error like this:

```python
da = xr.DataArray(
    np.arange(1, 4),
    coords=dict(time=np.arange('2000-01-01', '2000-01-04',
                               dtype=np.datetime64)),
    dims=['time'],
)
da_tznaive.sel(time=pd.Timestamp('2000-01-02T00:00:00Z'), method='nearest')
```

# But is datetime64 actually timezone-naive?

Actually, *no!*. The [NumPy docs make this
explicit](https://numpy.org/doc/stable/reference/arrays.datetime.html):

> NumPy does not store timezone information. For backwards compatibility, datetime64 still parses timezone offsets, which it handles by converting to UTC. This behaviour is deprecated and will raise an error in the future.
>
> …
>
> Datetimes are always stored based on POSIX time (though having a TAI mode which allows for accounting of leap-seconds is proposed), with an epoch of 1970-01-01T00:00Z. This means the supported dates are always a symmetric interval around the epoch, called “time span” in the table below.

So every datetime64 uses the UTC timezone, and the pandas behaviour causing the error above seems rather strange, because:

1. The `pandas.Timestamp` should not be converted into the deprecated `datetime64[ns, UTC]`, but instead to the equivalent `datetime64[ns]`

2. It should be possible to compare `datetime64[ns, UTC]` and `datetime64[ns]` because the second is implicitly also in UTC.

But whether or not we can blame this particular error on pandas, the problem of dealing with it safely is still on the xcube side.

# How should we proceed?

The world of time representations in general, and Python time representations in particular, is [notoriously messy](https://stackoverflow.com/questions/13703720/converting-between-datetime-timestamp-and-datetime64). I think that the most reliably way to make xcube robust against these errors is to check and sanitize inputs to `xarray.sel` at point of use, by making the selectors TZ-aware or TZ-naive if this is required to match the DataArray. This was already done for Issue #555 by introducing an `_ensure_time_compatible` function, which could form the basis of a more general solution:

1. Expand `_ensure_time_compatible` to check for all TZ-aware / TZ-naive incompatibilities (currently it just checks for TZ-aware selectors on `datetime64` arrays).

2. Apply `_ensure_time_compatible` wherever there's a potentially dangerous call to `DataArray.sel` (see below).

3. Make `_ensure_time_compatible` log a warning if it does have to modify a selector. This will help us to find fix the origin of the incompatibility rather than relying on on-the-fly sanitization.

Fortunately, I only found a few calls to `DataArray.sel` with a time selector in the xcube codebase, so step 2 should not be too time-consuming:

 - `core.select.select_temporal_subset`
 - `core.timeseries.get_time_series`
 - `cli.tile.tile`

pont-us:
Closes #605 .

Checklist:

* [x] AppVeyor CI passes


xcube/util/timeindex.py
```diff
@@ -0,0 +1,141 @@
...
+        except (TypeError, ValueError):
+            logger.warning("Can't determine indexer timezone, leaving it "
+                           "unmodified.")
+            return timeval
+
+    if _has_datetime64_time(var):
+        # pandas treats all datetime64 arrays as timezone-naive
+        array_timezone = None
+    elif hasattr(var.time.values[0], 'tzinfo'):
+        array_timezone = var.time.values[0].tzinfo
```
TonioF, line 113:
```suggestion
        array_timezone = var.time[0].values.tzinfo
```

xcube/util/timeindex.py
```diff
@@ -0,0 +1,141 @@
...
+    else:
+        logger.warning(
+            "Can't determine array timezone, leaving indexer unmodified.")
```
TonioF, line 116:
```suggestion
        logger.warning(
            "Can't determine array timezone, leaving indexer unmodified."
        )
```

TonioF (changes requested):
See comments, all minor.

xcube/util/timeindex.py
```diff
@@ -0,0 +1,141 @@
...
+
+    if hasattr(timeval, "tzinfo"):
+        timestamp = timeval
+        timeval_tzinfo = timeval.tzinfo
+    else:
+        try:
+            timestamp = pd.Timestamp(timeval)
+            timeval_tzinfo = timestamp.tzinfo
+        except (TypeError, ValueError):
+            logger.warning("Can't determine indexer timezone, leaving it "
```
AliceBalfanz, line 105: https://github.com/bc-org/best-practice/blob/main/coding/python.md#quotation-marks-for-string-literals

Is there a reason for double quotes?

AliceBalfanz: Also refers to all other occurrences of logger.

CHANGES.md
```diff
@@ -4,7 +4,8 @@
 
 ### Fixes
 
-
+* xcube now takes care to avoid indexing timezone-naive variables with
```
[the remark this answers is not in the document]
pont-us: "Prevents" is a bit too strong (in part because we can't guarantee that every datetime-like value will support naive/aware interconversion), but I'll compromise on "tries to prevent".

test/util/test_timeindex.py
```diff
@@ -0,0 +1,129 @@
...
+    # A DataArray with a datetime64 time dimension -- actually implicitly
+    # timezone-aware per the numpy docs, but treated as timzone-naive by
+    # pandas and xarray.
+    da_datetime64 = xr.DataArray(
+        np.arange(1, 4),
+        coords=dict(time=np.arange('2000-01-01', '2000-01-04',
+                                   dtype=np.datetime64)),
+        dims=['time'])
+
+    # Currently, pd.date_range seems to produce datetime64 co-ordinates, but
```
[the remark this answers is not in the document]
pont-us: Good catch, I'll specify the pandas version instead.

xcube/util/timeindex.py
pont-us: Rather surprisingly, this breaks the test, because indexing time like that causes xarray to automatically convert the value to a `datetime64` with a timezone (!). But I found a workaround, and will implement it here and in the line above:

``` python
var.time[0:1].values[0].tzinfo
```

Looks a little odd, but slicing the time rather than indexing it preserves the initial datatype, and avoids pulling the whole time array into memory.

xcube/util/timeindex.py
pont-us: In this particular case I could claim that it saves me from having to escape the single quote within the string, but actually I just got careless :). I will go through and singularize these and any other double quotes that sneaked in.

pont-us:
Commit a516f79 implements all suggested changes in the reviews from @AliceBalfanz and @TonioF .

TejasMorbagal (approved):

thomasstorm (approved):

xcube/util/timeindex.py
```diff
@@ -0,0 +1,142 @@
...
+import numpy as np
+import pandas as pd
+import xarray as xr
+import logging
+
+logger = logging.getLogger('xcube')
+
+
+def ensure_time_label_compatible(var: Union[xr.DataArray, xr.Dataset],
+                                 labels: Dict[str, Any]) -> Dict[str, Any]:
```
forman, line 39: I know, there are many code locations where we assume time coordinates and dimension is named "time". But I can tell, this is wrong. We once did the same mistake for "lat" and "lon" and we still suffer from this assumption, however, most of xcube code is now more tolerant. 

```suggestion
def ensure_time_label_compatible(var: Union[xr.DataArray, xr.Dataset],
                                 labels: Dict[str, Any],
                                 time_name: str = "time") -> Dict[str, Any]:
```

xcube/util/timeindex.py
```diff
@@ -0,0 +1,142 @@
...
+    in the var array, they are checked for compatibility. If they are
+    compatible, the original labels are returned. If not, an altered labels
+    dictionary is returned, in which the time key has been modified to be
+    compatible with the type of the 'time' dimension in the var array.
+
+    See the documentation for `ensure_time_index_compatible` for details on
+    the compatibility check.
+    """
+
+    if 'time' in labels and 'time' in var.dims:
```
forman, line 58:
```suggestion
    if time_name in labels and time_name in var.dims:
```

xcube/util/timeindex.py
```diff
@@ -0,0 +1,142 @@
...
+        return dict(labels,
+                    time=ensure_time_index_compatible(var, labels['time']))
```
forman, line 60:
```suggestion
                    time=ensure_time_index_compatible(var, labels, time_name))
```

forman: And so forth.

xcube/util/timeindex.py
```diff
@@ -0,0 +1,142 @@
...
+    else:
+        return labels
+
+
+def ensure_time_index_compatible(var: Union[xr.DataArray, xr.Dataset],
+                                 timeval: Any) -> Any:
```
forman, line 66: With only 3 characters more `timeval` turns into `time_value`, which is clearer and better fits to xcube variable naming convention (lower letter, snake case).

Please:
```suggestion
def ensure_time_index_compatible(var: Union[xr.DataArray, xr.Dataset],
                                 time_value: Any) -> Any:
```

xcube/util/timeindex.py
```diff
@@ -0,0 +1,142 @@
...
+    step field is included unchanged in the new slice.
+
+    The compatibility check consists of checking the timezone-awareness
+    of the variable and its indexer. If the variable is timezone-aware and
+    the indexer is timezone-naive, or vice versa, a new indexer is returned
+    whose timezone-awareness corresponds to that of the variable. Otherwise
+    the original indexer is returned.
+    """
+    
+    if isinstance(timeval, slice):
```
forman, line 83:
```suggestion
    timeval = labels[time_name]
    if isinstance(timeval, slice):
```

and so forth.

xcube/util/timeindex.py
```diff
@@ -0,0 +1,142 @@
...
+        return timeval
+
+
+def _has_datetime64_time(var: xr.DataArray) -> bool:
+    """Report whether var's time dimension has type datetime64
+
+    Assumes that a 'time' key is present in var.dims."""
+    return hasattr(var['time'], 'dtype') \
+           and hasattr(var['time'].dtype, 'type') \
+           and var['time'].dtype.type is np.datetime64
```
forman, line 142: Introduce and use parameter `time_name`.

xcube/util/timeindex.py
```diff
@@ -0,0 +1,142 @@
...
+        except (TypeError, ValueError):
+            logger.warning('Can\'t determine indexer timezone, leaving it '
+                           'unmodified.')
+            return timeval
+
+    if _has_datetime64_time(var):
+        # pandas treats all datetime64 arrays as timezone-naive
+        array_timezone = None
+    elif hasattr(var.time[0:1].values[0], 'tzinfo'):
+        array_timezone = var.time[0:1].values[0].tzinfo
```
forman, line 113: Extract variable for weird expression `var.time[0:1].values[0]` used twice and requires a comment.

xcube/util/timeindex.py
```diff
@@ -0,0 +1,142 @@
...
+                    time=ensure_time_index_compatible(var, labels['time']))
+    else:
+        return labels
+
+
+def ensure_time_index_compatible(var: Union[xr.DataArray, xr.Dataset],
+                                 timeval: Any) -> Any:
+    """Ensure that timeval is a valid time indexer for var
+
+    It is expected that the value of timeval will be a valid timestamp (i.e.
```
forman, line 69:
```suggestion
    It is expected that the value of *time_value* will be a valid timestamp (i.e.
```

xcube/util/timeindex.py
```diff
@@ -0,0 +1,142 @@
...
+        return dict(labels,
+                    time=ensure_time_index_compatible(var, labels['time']))
+    else:
+        return labels
+
+
+def ensure_time_index_compatible(var: Union[xr.DataArray, xr.Dataset],
+                                 timeval: Any) -> Any:
+    """Ensure that timeval is a valid time indexer for var
+
```
forman, line 68: Pont, you explain well, what the method does, could you please also explain why it is needed at all?

xcube/util/timeindex.py
forman: Explanation: parameter references in docstrings should be made italic.

xcube/core/tile.py
```diff
@@ -261,37 +262,15 @@ def _get_var_2d_array(var: xr.DataArray,
         if labels_are_indices:
             array = var.isel(**labels)
         else:
-            array = var.sel(method='nearest', **labels)
+            safe_labels = ensure_time_label_compatible(var, labels)
+            array = var.sel(method='nearest', **safe_labels)
```
forman, line 266: `xcube/core/tile.py` will be deprecated. For xcube server NG, a lot of code changes have been introduced. After merging this PR, please update also https://github.com/dcs4cop/xcube/pull/677.

xcube/core/tile.py
```diff
@@ -404,6 +383,6 @@ def to_datetime(datetime_str: str, dim_var: xr.DataArray):
                                  f' value for dimension {dim!r}') from e
 
     if var is not None:
-        return _ensure_time_compatible(var, parsed_labels)
+        return ensure_time_label_compatible(var, parsed_labels)
```
forman, line 386: `xcube/core/tile.py` will be deprecated. For xcube server NG, a lot of code changes have been introduced. After merging this PR, please update also https://github.com/dcs4cop/xcube/pull/677.

forman (changes requested):
Wow, thanks, quite complicated stuff! 

* Please see my comments and follow suggestions.
* Please check the runtime impact. Does this introduce performance issues?
* `xcube/core/tile.py` will be deprecated. For xcube server NG, a lot of code changes have been introduced. After merging this PR, please update also https://github.com/dcs4cop/xcube/pull/677.

xcube/util/timeindex.py
pont-us: I don't understand this suggestion. In this scope, `timeval` is already defined and `labels` is not bound, so we can't reassign it like this.

xcube/util/timeindex.py
pont-us: Done.

xcube/util/timeindex.py
pont-us: Done.

xcube/util/timeindex.py
pont-us: Done.

xcube/util/timeindex.py
pont-us: Done.

pont-us:
Norman has now OK'd the performance figures above. This was the last outstanding item, so I'm merging.

PR #696, opened 2022-06-03
Added maintenance endpoint to server

forman:
@pont-us , @TonioF more on this via a different channel.
FYI @AliceBalfanz.


pont-us (approved):
Looks good!

PR #700, opened 2022-06-17
Endpoints for xcube Server NG

forman:


test/server/mocks.py
```diff
@@ -20,7 +20,7 @@
 #  DEALINGS IN THE SOFTWARE.
 
 from typing import Optional, Sequence, Tuple, Dict, Any, Union, Callable, \
-    Awaitable, List
+    Awaitable, List, Mapping
```
thomasstorm, line 23: I'm pretty sure you told me that we only have a single import per line!

xcube/server/api.py
```diff
@@ -48,9 +49,10 @@
     Dict[str, "JSON"],
 ]
 
-ServerConfig = Mapping[str, Any]
+ServerConfigObject = Mapping[str, Any]
+ServerConfig = ServerConfigObject
```
thomasstorm, line 53: I don't understand why this is necessary, but you introduced it for a reason, I'm sure.

xcube/server/api.py
```diff
@@ -632,10 +659,21 @@ class ApiError(Exception):
...
+    BadRequest: Type["_DerivedApiError"]
+    Unauthorized: Type["_DerivedApiError"]
+    Forbidden: Type["_DerivedApiError"]
+    NotFound: Type["_DerivedApiError"]
+    MethodNotAllowed: Type["_DerivedApiError"]
+    Conflict: Type["_DerivedApiError"]
+    Gone: Type["_DerivedApiError"]
+    InternalServerError: Type["_DerivedApiError"]
+    NotImplemented: Type["_DerivedApiError"]
+    InvalidServerConfig: Type["_DerivedApiError"]
```
thomasstorm, line 675: Why are they all of ``Type["_DerivedApiError"]``? This seems wrong to me, because of line 747ff, and ``_BadRequest`` is not a ``_DerivedApiError``.

xcube/server/api.py
```diff
@@ -632,10 +659,21 @@ class ApiError(Exception):
     """
 
     def __init__(self,
-                 status_code: int,
+                 status_code: int = 500,
```
thomasstorm, line 662: Seems a bit random to set 500 as default. I don't see a good reason for this.

xcube/webapi/context.py
```diff
@@ -461,11 +461,11 @@ def get_dataset_configs_from_stores(self, data_store_pool: DataStorePool) \
...
-        # Just for testing:
-        debug_file = 'all_dataset_configs.json'
-        with open(debug_file, 'w') as stream:
-            json.dump(all_dataset_configs, stream)
-            LOG.debug(f'Wrote file {debug_file!r}')
+        # # Just for testing:
+        # debug_file = 'all_dataset_configs.json'
+        # with open(debug_file, 'w') as stream:
+        #     json.dump(all_dataset_configs, stream)
+        #     LOG.debug(f'Wrote file {debug_file!r}')
```
thomasstorm, line 468: Maybe then simply remove? We got version control if you want it back :)

xcube/webapi/datasets/routes.py
```diff
@@ -19,20 +19,119 @@
...
 
-# noinspection PyAbstractClass
-@api.route("/datasets")
+@api.route('/datasets')
 class DatasetsHandler(ApiHandler[DatasetsContext]):
+    """List the published datasets."""
+
     @api.operation(
-        operationId='getDatasets',
-        summary='Get all the datasets published by this service.'
```
thomasstorm, line 33: Why remove the summary?

xcube/webapi/meta/routes.py
```diff
@@ -0,0 +1,235 @@
...
+                                         'openapi.html').decode('utf-8')
+        self.response.finish(Template(html_template).substitute(
+            open_api_url=self.request.url_for_path('openapi.json')
+        ))
+
+
+@api.route("/openapi.json")
+class OpenApiJsonHandler(ApiHandler):
+    @api.operation(
+        operation_id='getOpenApiHtml',
```
thomasstorm, line 82:
```suggestion
        operation_id='getOpenApiJson',
```

thomasstorm (changes requested):
Minor stuff only!

xcube/webapi/context.py
forman: No, see, it has a comment. I often need that.

xcube/webapi/datasets/routes.py
forman: Have a look at the generated page. This is how Swagger works.

forman: It displays the summary just next to the route path which is very handy.

xcube/server/api.py
forman: Yes, this significantly improves the readability of the code when used in type annotations. It adds semantics to arguments and return types. Common practice in Python. Btw, in generated API docs, the type aliases will go away.

xcube/webapi/datasets/routes.py
thomasstorm: Ok, but why keep the summary e.g. in line 86?

xcube/webapi/context.py
forman: Then I simple un-comment it. Our rule is that commented-out code can stay if it has a header comment.

xcube/server/api.py
forman: By accident! I'll fix that.

xcube/webapi/datasets/routes.py
forman: Aaah, _remove_! I read _add_. I'll fix it.

xcube/server/api.py
forman: I use the dummy type `_DerivedApiError` because I cannot use `Type["ApiError"]` and this because it has status code as first positional argument and then expressions such as `raise ApiError.BadRequest()` would be shown as error by the IDE's type checker.

xcube/webapi/datasets/routes.py
forman: Fixed.

xcube/server/api.py
forman: Fixed.

thomasstorm (approved):

PR #703, opened 2022-06-24
Fixed determination of bbox from metadata

TonioF:
Minor PR that fixes the determination of a bounding box from metadata attributes.


forman (approved):
Oh, oh. Thanks!

PR #721, opened 2022-09-01
Fixed various issues with the filessystem-based data stores.

forman:
* The filesystem-based data stores for the "s3", "file", and "memory" protocols can now provide `xr.Dataset` instances from image pyramids formats, i.e. the `levels` and `geotiff` formats.
* Fixed various other code issues with the filessystem-based data stores.
* Improved test coverage.

Checklist:

* [x] AppVeyor CI passes

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,10 @@
 
 ### Fixes
 
+* The filesystem-based data stores for the "s3", "file", and "memory"
+  protocols can now provide `xr.Dataset` instances from image pyramids
+  formats, i.e. the `levels` and `geotiff` formats.
+
 
 ## Changes in 0.12.0
 
--- a/test/core/store/fs/test_registry.py
+++ b/test/core/store/fs/test_registry.py
@@ -3,10 +3,11 @@ import shutil
 import unittest
 import warnings
 from abc import ABC, abstractmethod
-from typing import Any, Callable, Dict, Optional, Type, Union
+from typing import Any, Callable, Dict, Optional, Set, Type, Union
 
 import fsspec
 import numpy as np
+import pytest
 import xarray as xr
 
 import xcube.core.mldataset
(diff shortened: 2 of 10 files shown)


TejasMorbagal (approved):

xcube/core/store/fs/impl/mldataset.py
```diff
@@ -225,10 +236,34 @@ def write_data(self,
         return data_id
 
 
-_MIN_CACHE_SIZE = 1024 * 1024  # 1 MiB
+class DatasetLevelsFsDataAccessor(MultiLevelDatasetLevelsFsDataAccessor):
+    """
+    Opener/writer extension name "dataset:levels:<protocol>".
+    """
```
TonioF, line 242: So, if I see this right, this accessor allows to (a) open datasets with multiple levels as if they only had one and (b) save a dataset with a single level as a multileveldataset with multiple levels. If this is so, I'd like it to be documented more. If it is not so, documentation is even more important.

xcube/core/store/fs/store.py
```diff
@@ -187,12 +189,25 @@ def get_data_store_params_schema(cls) -> JsonObjectSchema:
...
+            for types_tuple in _FORMAT_TO_DATA_TYPE_ALIASES.values()
+            for data_type in types_tuple
+        ))
 
     def get_data_types_for_data(self, data_id: str) -> Tuple[str, ...]:
         self._assert_valid_data_id(data_id)
-        data_type_alias, _, _ = self._guess_accessor_id_parts(data_id)
-        return data_type_alias,
+        data_type_alias, format_id, protocol = \
+            self._guess_accessor_id_parts(data_id)
```
TonioF, line 201: This doesn't look right to me. self._guess_accessor_id_parts could be able to return more than one data_type_alias, however, just the first one is returned. We are not necessarily returning all data types for data in this method.

TonioF (changes requested):
Please consider in particular my last comment.

xcube/core/store/fs/store.py
forman: This is the logic that is currently implemented and it has been like this for a long time. Please post a report, if you found an issue and describe the problem that you observe.

CHANGES.md
```diff
@@ -11,6 +11,10 @@
 
 ### Fixes
 
+* The filesystem-based data stores for the "s3", "file", and "memory"
+  protocols can now provide `xr.Dataset` instances from image pyramids
+  formats, i.e. the `levels` and `geotiff` formats.
+
```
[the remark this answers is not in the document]
forman: No, "levels" format (= opener) has formerly reported to be able to open datasets, but it didn't.

xcube/core/store/fs/impl/mldataset.py
forman: This is documented in the `plugin.py`. To view, please use CLI `xcube io opener list`.

forman: Note, it is not the intention of this PR to add docs or to enhance the framework. This is just about fixing code and testing issues.

TonioF: > This is documented in the `plugin.py`. To view, please use CLI `xcube io opener list`.

Okay. I'm not 100% happy with the phrasing, but this is clearly outside the scope of this PR.

xcube/core/store/fs/store.py
```diff
@@ -556,14 +580,16 @@ def _guess_accessor_id_parts(self, data_id: str, require=True) \
...
+        format_id = _FILENAME_EXT_TO_FORMAT.get(ext)
+        if format_id is not None:
+            data_type_aliases = _FORMAT_TO_DATA_TYPE_ALIASES.get(format_id)
+        if data_type_aliases is None or format_id is None:
             if require:
                 raise DataStoreError(f'Cannot determine data type for '
                                      f' data resource {data_id!r}')
             return None
-        return data_type_alias, format_name, self.protocol
+        return data_type_aliases[0], format_id, self.protocol
```
TonioF, line 592: This is the part where we arbitrarily pick one data_type_alias (the first one). Depending on how this method has been used, it might introduce problems if we omit other data type aliases.

xcube/core/store/fs/store.py
TonioF: No, this is about a newly introduced change. See my new comment in _guess_accessor_id_parts.

CHANGES.md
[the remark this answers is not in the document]
TonioF: So, the error was that the docs stated that a functionality existed which did not exist? I'd still consider this an enhancement (at the same time, it is nitpicking, so this really shouldn't stop us from accepting this pr).

xcube/core/store/fs/store.py
forman: No, the implementation defines the first alias to be the one. This is stored in constants defined at the top of the file. Secondly, they are just aliases, so they all point to the same data type.

forman: ...and so far, tests are green too.

test/core/store/fs/test_registry.py
```diff
@@ -137,34 +138,61 @@ def test_mldataset_levels(self):
...
+
+    def test_dataset_levels(self):
+        data_store = self.create_data_store()
+        self._assert_dataset_supported(
+            data_store,
+            filename_ext='.levels',
+            requested_dtype_alias='dataset',
+            expected_dtype_aliases={'mldataset', 'dataset'},
+            expected_return_type=xr.Dataset,
+            expected_descriptor_type=None
```
TonioF, line 169: So, we can use the store to open a dataset, but not to get a descriptor? This does not seem right.

TonioF (changes requested):

test/core/store/fs/test_registry.py
forman: `expected_descriptor_type=None` means we do not want to test for the descriptor, because test code would be too confusing. It does not mean, the descriptor must not exist.

xcube/core/store/fs/store.py
TonioF: They stop running green if we change the order of aliases in _FORMAT_TO_DATA_TYPE_ALIASES. However, I guess it is okay if we say that the first data type is the primary one,

test/core/store/fs/test_registry.py
TonioF: I'm sorry, but to me it looks like we are configuring the tests in a way that they run green. It should work if we specify expected_descriptor_type=DatasetDescriptor , just like in the other tests.

forman: >> It should work if we specify expected_descriptor_type=DatasetDescriptor , just like in the other tests.

No, because the descriptor is retrieved using `search_data(expected_data_type)`. But this will not return a descriptor for dataset (yet), because it is currently implemented to return MultiLevelDatasetDescriptorsa only.

Again, this is not part of this PR.

TonioF: Okay, then please go ahead and merge.

PR #729, opened 2022-09-15
Fix broken JSON encoding

### LINKED ISSUE #688: xcube to write sparse zarrs
When I create an xcube dataset in zarr, all chunks are written to disk, even ones which are all NaNs. after writing the cube, I apply xcube prune to get rid of empty chunks to save disk space. zarr introduced in a recent release https://zarr.readthedocs.io/en/stable/release.html#release-2-11-0 that nan chunks are not written to disk. xcube should use that right away to save space and (user) time. 

To reproduce: 
I have an xcube env with zarr version 2.11.3 

1. I write a cube where I know whole chunks filled with Nans exist. 
2. I make a list of all files of the zarr directory
3. I apply xcube prune to the newly created zarr
4. make a list of the zarr directory again and compare it to the previous list. There I see, that xcube does not actively use zarrs enhancement yet, because the list after pruning is way shorter than the initial one.

forman:
* The xcube Python environment is now requiring  `xarray >= 2022.6` and `zarr >= 2.11` to ensure sparse Zarr datasets can be written using `dataset.to_zarr(store)`. (#688)
* Added new module `xcube.util.jsonencoder` that offers the class `NumpyJSONEncoder` used to serialize numpy-like scalar values to JSON. It also offers the function `to_json_value()` to convert Python objects into JSON-serializable versions. The new functionality is required to ensure dataset attributes that are JSON-serializable. For example, the latest version of the `rioxarray` package generates a `_FillValue` attribute with datatype `np.uint8`. 

Closes #688.

**EDIT**

**It is likely that other places where we `json.dump()` are affected too!** In particular, where numeric metadata attributes are serialized such as `_FillValue`, `add_offset`, `valid_max`, etc.

Checklist:

* [x] AppVeyor CI passes

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -32,13 +32,13 @@
   `xarray >= 2022.6` and `zarr >= 2.11` to ensure sparse 
   Zarr datasets can be written using `dataset.to_zarr(store)`. (#688)
 
-* Added new module `xcube.util.jsonencoder` that 
-  offers class `NumpyJSONEncoder` that  used to serialize 
-  numpy-like scalar values to JSON. It also offers the function 
-  `to_json_value()` to convert Python objects into JSON-serializable 
-  versions. The new functionality is required to ensure dataset
-  attributes that are JSON-serializable. For example, latest
-  `rioxarray` generates a `_FillValue` attribute with datatype `np.uint8`. 
+* Added new module `xcube.util.jsonencoder` that offers the class 
+  `NumpyJSONEncoder` used to serialize numpy-like scalar values to JSON. 
+  It also offers the function `to_json_value()` to convert Python objects 
+  into JSON-serializable versions. The new functionality is required 
+  to ensure dataset attributes that are JSON-serializable. For example,
+  the latest version of the `rioxarray` package generates a `_FillValue` 
+  attribute with datatype `np.uint8`. 
 
 ### Fixes
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -28,6 +28,13 @@
   is used to select dataset labels along a given dimension using
   user-defined predicate functions.
 
+* The xcube Python environment is now requiring 
+  `xarray >= 2022.6` and `zarr >= 2.11` to ensure sparse 
+  Zarr datasets can be written using `dataset.to_zarr(store)`. (#688)
+
+* Added new class `xcube.util.jsonencoder.NumpyJSONEncoder` that 
+  is used to serialize numpy-like scalar values to JSON.
+
 ### Fixes
 
 * The filesystem-based data stores for the "s3", "file", and "memory"
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -34,6 +34,8 @@
 
 * Added new class `xcube.util.jsonencoder.NumpyJSONEncoder` that 
   is used to serialize numpy-like scalar values to JSON.
+  The new encoder is required because `xcube io dump` was broken 
+  due to the updated env.
 
 ### Fixes
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -32,10 +32,13 @@
   `xarray >= 2022.6` and `zarr >= 2.11` to ensure sparse 
   Zarr datasets can be written using `dataset.to_zarr(store)`. (#688)
 
-* Added new class `xcube.util.jsonencoder.NumpyJSONEncoder` that 
-  is used to serialize numpy-like scalar values to JSON.
-  The new encoder is required because `xcube io dump` was broken 
-  due to the updated env.
+* Added new module `xcube.util.jsonencoder` that 
+  offers class `NumpyJSONEncoder` that  used to serialize 
+  numpy-like scalar values to JSON. It also offers the function 
+  `to_json_value()` to convert Python objects into JSON-serializable 
+  versions. The new functionality is required to ensure dataset
+  attributes that are JSON-serializable. For example, latest
+  `rioxarray` generates a `_FillValue` attribute with datatype `np.uint8`. 
 
 ### Fixes
 
--- a/test/util/test_jsonencoder.py
+++ b/test/util/test_jsonencoder.py
@@ -1,0 +1,85 @@
+# The MIT License (MIT)
+# Copyright (c) 2022 by the xcube development team and contributors
+#
+# Permission is hereby granted, free of charge, to any person obtaining a
+# copy of this software and associated documentation files (the "Software"),
+# to deal in the Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish, distribute, sublicense,
+# and/or sell copies of the Software, and to permit persons to whom the
+# Software is furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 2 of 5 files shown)


forman:
FYI @AliceBalfanz @thomasstorm @TejasMorbagal

test/util/test_jsonencoder.py
```diff
@@ -0,0 +1,85 @@
...
+        "np_uint32": np.uint32(6),
+        "np_int64": np.int64(7),
+        "np_uint64": np.uint64(8),
+        "np_float32": np.float32(9.1),
+        "np_float64": np.float64(9.2),
+        "py_bool": True,
+        "py_int": 11,
+        "py_float": 12.3,
+        "py_str": "Hallo",
+        "py_null": None,
```
pont-us, line 48: Quotation marks should be single per current xcube dev guide (or use `dict(...)` to avoid quotes entirely in the keys).

test/util/test_jsonencoder.py
```diff
@@ -0,0 +1,85 @@
...
+                "np_uint32": int(np.uint32(6)),
+                "np_int64": int(np.int64(7)),
+                "np_uint64": int(np.uint64(8)),
+                "np_float32": float(np.float32(9.1)),
+                "np_float64": float(np.float64(9.2)),
+                "py_bool": True,
+                "py_int": 11,
+                "py_float": 12.3,
+                "py_str": "Hallo",
+                "py_null": None,
```
pont-us, line 82: Again, single quote marks / `dict(...)` preferred.

xcube/util/jsonencoder.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class NumpyJSONEncoder(json.JSONEncoder):
+    """A JSON encoder that converts numpy-like
+    scalars into corresponding serializable Python objects.
+    """
+
+    def default(self, obj):
+        if hasattr(obj, 'dtype') and hasattr(obj, 'ndim'):
+            if obj.ndim == 0:
```
pont-us, line 34: I'd prefer to chain this condition onto the `if` above with another `and` rather than adding another indentation level. Python guarantees short-circuit evaluation in this case, so we're safe from `AttributeError`s.

xcube/util/jsonencoder.py
```diff
@@ -0,0 +1,45 @@
...
+                # For time being just handle scalars.
+                if np.issubdtype(obj.dtype, np.bool):
+                    return bool(obj)
+                if np.issubdtype(obj.dtype, np.integer):
+                    return int(obj)
+                elif np.issubdtype(obj.dtype, np.floating):
+                    return float(obj)
```
pont-us, line 41: As far as I can see, `if` and `elif` are functionally equivalent for these conditions, since they're mutually exclusive and every clause exits the function. I don't have a stylistic preference for one or the other, but I would prefer them all to be the same :).

xcube/util/jsonencoder.py
```diff
@@ -0,0 +1,45 @@
...
+                else:
+                    return str(obj)
```
pont-us, line 43: I'm not sure which types are covered by this case, but I guess it can't hurt.

pont-us (changes requested):
Looks good and tests pass locally. I made some nitpicking suggestions, but I don't insist on any of them :).

xcube/util/jsonencoder.py
forman: No - see my code comment.

test/util/test_jsonencoder.py
forman: > Quotation marks should be single

Right.

test/util/test_jsonencoder.py
forman: Right.

xcube/util/jsonencoder.py
forman: Right.

xcube/util/jsonencoder.py
forman: I just discovered, that my implementation is possibly wrong because `JSONEncoder.default(self, obj)` raises `TypeError` by default. Adding extra test...

forman:
FYI @TonioF, I need to merge today

xcube/util/jsonencoder.py
```diff
@@ -0,0 +1,113 @@
...
+_PRIMITIVE_JSON_TYPES = {
+    type(None),
+    bool,
+    int,
+    float,
+    str
+}
+
+
+def to_json_value(obj: Any) -> Union[Any, JsonValue]:
```
forman, line 53:
```suggestion
def to_json_value(obj: Any) -> JsonValue:
```

xcube/util/jsonencoder.py
```diff
@@ -0,0 +1,113 @@
...
+                return bool(obj)
+            elif np.issubdtype(obj.dtype, np.integer):
+                return int(obj)
+            elif np.issubdtype(obj.dtype, np.floating):
+                return float(obj)
+            elif np.issubdtype(obj.dtype, np.str):
+                return str(obj)
+        else:
+            return [_convert_default(item) for item in obj]
+    # We may handle other non-sJSON-serializable datatypes here
```
forman, line 112:
```suggestion
    # We may handle other non-JSON-serializable datatypes here
```

test/util/test_jsonencoder.py
```diff
@@ -0,0 +1,172 @@
...
+
+EXPECTED_PY_OUTPUT = {
+    'py_bool': True,
+    'py_int': 11,
+    'py_float': 12.3,
+    'py_str': 'Hallo',
+    'py_null': None,
+    'py_list': [1, 2, 3],
+    'py_dict': {'x': 'A', 'y': 'B'},
+}
```
forman, line 70:
```suggestion
EXPECTED_PY_OUTPUT = {**PY_INPUT}
```

CHANGES.md
```diff
@@ -28,6 +28,15 @@
...
   user-defined predicate functions.
 
+* The xcube Python environment is now requiring 
+  `xarray >= 2022.6` and `zarr >= 2.11` to ensure sparse 
+  Zarr datasets can be written using `dataset.to_zarr(store)`. (#688)
+
+* Added new class `xcube.util.jsonencoder.NumpyJSONEncoder` that 
+  is used to serialize numpy-like scalar values to JSON.
+  The new encoder is required because `xcube io dump` was broken 
+  due to the updated env.
```
forman, line 38:
```suggestion
* Added new module `xcube.util.jsonencoder` that 
  offers class `NumpyJSONEncoder` that  used to serialize 
  numpy-like scalar values to JSON. It also offers the function 
  `to_json_value()` to convert Python objects into JSON-serializable 
  versions. The new functionality is required to ensure dataset
  attributes that are JSON-serializable. For example, latest
  `rioxarray` generates a `_FillValue` attribute with datatype `np.uint8`. 
```

xcube/util/jsonencoder.py
```diff
@@ -0,0 +1,113 @@
...
+    if obj_type in _PRIMITIVE_JSON_TYPES:
+        return obj
+
+    for t in _PRIMITIVE_JSON_TYPES:
+        if isinstance(obj, t):
+            return t(obj)
+
+    if obj_type is dict:
+        converted_obj = {k: to_json_value(v) for k, v in obj.items()}
+        if any(converted_obj[k] is not obj[k] for k in obj.keys()):
```
pont-us, line 76: Keys can be non-serializable too, so `converted_obj = {to_json_value(k): to_json_value(v)...` would be useful. And then of course we'd also need to check whether the keys are identical in the `if` statement on the next line.

No need to implement now if it's working for our current needs, but a TODO comment might be appropriate.

pont-us (changes requested):
Tests pass, and code makes sense to me.

My one requested change just involves adding a TODO comment for a possible useful expansion, so ignore it if you're in a hurry.

xcube/util/jsonencoder.py
forman: `converted_obj = {k: to_json_value(v) for k, v in obj.items()}` is wrong because keys must be strings in JSON. I agree to change that, but later.

PR #730, opened 2022-09-16
Fix identification of grid mapping

forman:
This PR fixes the following issue in current master:

When running xcube serve with the updated xcube env (xarray >= 2022.6, zarr >= 2.11)

    $ xcube serve -v -c examples/serve/demo/config.yml

and do `GET http://localhost:8080/datasets?details=1`

I receive

```json
{
  "error": {
    "code": 500,
    "message": "Internal Server Error",
    "traceback": [
      "Traceback (most recent call last):\n",
	  ...
      "ValueError: cannot find any grid mapping in dataset\n"
    ]
  }
}
```

For some strange reason, `lon_bnds` and `lat_bnds` are selected as potential spatial coordinates over `lon` and `lat`.
That never happened before, but the code indeed does not exclude that possibility.
Later on in the process, they are rejected. Finally we have no gridmapping. 

Unfortunately I cannot easily write a test case, because the issue seems to occur with the server's demo configuration only.

Checklist:

* [ ] ~Changes documented in `CHANGES.md`~
* [ ] AppVeyor CI passes

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -34,6 +34,8 @@
 
 * Added new class `xcube.util.jsonencoder.NumpyJSONEncoder` that 
   is used to serialize numpy-like scalar values to JSON.
+  The new encoder is required because `xcube io dump` was broken 
+  due to the updated env.
 
 ### Fixes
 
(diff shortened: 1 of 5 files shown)


thomasstorm (approved):

pont-us (approved):
Code looks correct, and fixes the error you describe when I run `xcube serve ...` locally. Approved.

PR #737, opened 2022-09-22
Enhance xcube Server docs

forman:
Describing how display value ranges and colorbars are determined by xcube server.

This will be the first of a series of doc updates regarding xcube Server NG.


AliceBalfanz (approved):
Thanks for adding to the docs!

PR #743, opened 2022-10-06
Filesystem data stores to filter content

### LINKED ISSUE #585: Allow explicit configuration of datasets served by the S3 store

Consider a situation where we want a store to serve data from Zarrs from a very large S3 bucket (millions of files, petabytes of data; e.g. Eodata on Creodias). The usual store creation parameters, `root` and `max_depth`, are insufficient, because:

 - The Zarrs we want to serve are not guaranteed to be grouped under a common prefix
 - A search for Zarrs in a bucket of this size, even starting from the longest common prefix, would take a very long time
 - There may be Zarrs that we wish to exclude from the store, for instance because of known problems with data or compatibility
 - The paths of the Zarrs in the store may have lengthy prefixes and names, and may therefore be inconvenient for users to work with

I would like to be able to do something like the following:

```python
store = new_data_store('s3', root='some-bucket',
            serve_only={
                'nice-name-1': 'long/messy/path/long-and-messy-name.zarr',
                'nice-name-2': 'some/other/long/path/another-long-name.zarr',
                'nice-name-3': 'i/think/you/get/the-idea.zarr'
            })
```

Supplying the `serve_only` parameter disables automatic data discovery, and serves only the datasets specified in the dictionary values, presenting them to the user under the aliases given by the dictionary keys.

 - Creating a separate store for each dataset: inconvenient (especially for large numbers of datasets), doesn't allow aliasing, doesn't allow exclusion of unwanted datasets with the same prefix as wanted datasets
 - Include/exclude list parameters: doesn't allow aliasing; doesn't necessarily disable automatic dataset discovery
 - Implement an external S3 ‘facade’, using the original bucket as a back-end and serving selected Zarrs under selected aliases: more work and more points of failure than implementing the suggested feature directly in xcube.

forman:
The xcube filesystem data stores such as "file", "s3", "memory" can now filter the data identifiers reported by `get_data_ids()`. (#585)

For this purpose, the data stores now accept two new optional keywords which both can take the form of a wildcard pattern or a sequence of wildcard patterns:

1. `excludes`: if given and if any pattern matches the identifier, the identifier is not reported. 
2. `includes`: if not given or if any pattern matches the identifier, the identifier is reported.

Closes #585

Checklist:

* [x] AppVeyor CI passes


pont-us (approved):
Tests pass and code makes sense – approving.

TonioF (approved):
I don't see the necessities of the FileFsStoreSubsetTest, the S3FsStoreSubsetTest and the change in config.py for this change, but these changes don't seem to break anything. The includes/excludes parts work fine.

forman:
FYI @AliceBalfanz @TejasMorbagal @thomasstorm

PR #747, opened 2022-10-07
updated environment

TonioF:
Updated environment, as after latest implementations xcube does not work with python 3.8 or matplotlib < 3.6.0

Checklist:
* [x] AppVeyor CI passes


TonioF:
Removed part on upgrading matplotlib version as I was not able to reproduce the problem. I added an issue describing the problem with the python version: #760

forman:
#760 has been fixed by #770

PR #753, opened 2022-10-18
Remove outdated xcube server resources

forman:
Removed outdated example resources from `examples/serve/demo`.


AliceBalfanz:
I wish, you would have had a look at the already existing PR: https://github.com/dcs4cop/xcube/pull/751 which does the same.

AliceBalfanz (approved):

PR #771, opened 2022-12-09
Providing compatibility with Python 3.10

### LINKED ISSUE #583: xcube is not compatible with Python 3.10
Since the release of Python 3.10.2 on January 14th, 2022, xcube builds fail with errors like: 

ImportError: cannot import name 'MutableMapping' from 'collections'

This is because abstract base classes have been moved from Python's 'collections' package to 'collections.abc'. While there are a few instances in the xcube code where the import needs to be adjusted, this issue also affects the pydap package, which is a transitive dependency of xarray and is beyond our fixing scope.

Also, the relevant change has already been included in Python 3.10, for some reason, Python 3.9 was used for xcube builds until the release of Python 3.10.2.

forman:
Majority of issues come from outdated `moto`. Now, with current `moto >=4`, we run into https://github.com/aio-libs/aiobotocore/issues/755 with no obvious solution.

Closes #583

Checklist:

* [ ] Changes documented in `CHANGES.md`  (will add after merge, so we don't need CI once more)
* [x] AppVeyor CI passes


TonioF (approved):
Looks good. Do these changes change anything about this: https://github.com/dcs4cop/xcube/blob/cc3690c6ad889121b66511ad7a08c384a1ba5ac2/test/core/store/fs/test_subset.py#L164 , though?

environment.yml
```diff
@@ -4,7 +4,7 @@ channels:
   - defaults
 dependencies:
   # Python
-  - python >=3.8,<3.10
+  - python >=3.8
```
pont-us, line 7:
```suggestion
  - python >=3.8,<3.11
```

Currently, creating an xcube environment with the one-sided `>=3.8` limit installs Python 3.10, I assume because something in the dependencies is currently requiring `python <3.11`. That dependency will one day suddenly start supporting Python 3.11 and new xcube environments will start getting Python 3.11 by default with no changes on our side. That seems dangerous, so I suggest including the upper limit, but I leave the decision to you.

pont-us (approved):
Approved, with one optional suggestion.

environment.yml
TonioF: > something in the dependencies is currently requiring `python <3.11`

It would be interesting to know what that something is.

pont-us: > It would be interesting to know what that something is.

Agreed -- I considered investigating, but decided against it since explorations of conda dependency structures can get quite time-consuming.

forman: So you know, Python 3.11 is _not_ working?

forman:
> Looks good. Do these changes change anything about this:

I don't know. But we should find out after merging.

environment.yml
forman: If not, we stay with `python >=3.8` and fix any issues with `3.11` when they occur.

pont-us: I don't know that Python 3.11 is *not* working, but I also don't know how to test that it *will* work. So if you want to take the chance we can leave it without the upper limit :)

PR #775, opened 2022-12-30
xcube JupyterLab integration

forman:
This PR is one step towards running xcube server/viewer from within Jupyter Notebooks.

This PR provides a new _interactive xcube server_ that can be launched programmatically from interactive shells such as Jupyter Notebooks. The interactive server can listen to any user-provided port. Users can add to the server, remove them, and list the currently served datasets. 

**Important:** Use a build of xcube Viewer PR https://github.com/dcs4cop/xcube-viewer/pull/264.

A number of enhancements have been added to `xcube serve` in order to allow for or support integration with JupyterLab:

* added path arguments that identify datasets to be served
* added new endpoint `/viewer/` that serves xcube Viewer, recognizing env var `XCUBE_VIEWER_PATH`
* added `--open-viewer` option that can open xcube Viewer after the server started
* added `--revprefix PREFIX` option that is used to set the reverse URLs returned by the server

Note to test the JupyterLab integration, the JupyterLab extension [`xcube-jl-ext`](https://github.com/dcs4cop/xcube-jl-ext) must be installed (Python package `xcube_jl_ext`).
To test the integration with a remote JupyterLab, also the Jupyter server extension `jupyter-server-proxy`must be installed too.
 
Checklist:

* [x] AppVeyor CI passes


thomasstorm (approved):
Just a few minor suggestions, otherwise I'm happy!

CHANGES.md
```diff
@@ -95,7 +114,9 @@
 
 ### Fixes
 
-* xcube serve correctly resolves relative paths to datasets (#758)
+* Relative local filesystem paths to datasets are now correctly resolved 
+  against the base directory of the xcube Server's configuration, i.e.
```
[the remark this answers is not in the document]
forman: "using the```" ?

AliceBalfanz (changes requested):
> added --open-viewer option that can open xcube Viewer after the server started

This only opens a new browser window of the default browser - but empty, not with the viewer url.

xcube/webapi/viewer/viewer.py
```diff
@@ -0,0 +1,182 @@
+# The MIT License (MIT)
+# Copyright (c) 2022 by the xcube team and contributors
```
AliceBalfanz, line 2:
```suggestion
# Copyright (c) 2023 by the xcube team and contributors
```

test/webapi/viewer/test_routes.py
```diff
@@ -0,0 +1,70 @@
+# The MIT License (MIT)
+# Copyright (c) 2022 by the xcube team and contributors
```
AliceBalfanz, line 2:
```suggestion
# Copyright (c) 2023 by the xcube team and contributors
```

AliceBalfanz (commented):
When I use the parameter `--prefix xcube/api/latest`
the output of 
http://localhost:8080/xcube/api/latest/

looks as expected - but if I add the viewer parameter, the datasets are not shown:

image

without passing a --prefix parameter this works fine:
image

forman:
> ... but if I add the viewer parameter, the datasets are not shown:

Unfortunately, this cannot easily be achieved because the viewer uses it's default server API configuration, which is at `http://localhost:8080`.

We need to make available an alternate viewer configuration via `http://localhost:8080/viewer/config/config.json` but this isn't yet implemented. After 0.13 release, let's create a feature request for that:

- xcube serve --config <PATH> ..., --> allow PATH to be an URI, e.g. pointing into S3 
- add server API endpoint /viewer/config/ to serve static viewer config resources
- allow xcube server config to also provide viewer config 


Then, we no longer need

- s3fuse deployments
- viewer deployments (because xcube server endpoint /viewer provides it already)

AliceBalfanz (approved):
Thanks for fixing the --open-viewer param

PR #782, opened 2023-01-13
fixed github action workflow

TonioF:
Fixes the github action workflow. I have to confess I don't understand why the workflow broke (all I know is that miniconda was adamant about using python 3.10). I also tried other combinations of miniconda and setup-miniconda, but this is the one that works.
I propose to go for a new configuration as soon as xcube runs on python 3.10, in the meantime, this will work.

### DIFF
--- a/.github/workflows/xcube_workflow.yaml
+++ b/.github/workflows/xcube_workflow.yaml
@@ -18,7 +18,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       # Setup miniconda build env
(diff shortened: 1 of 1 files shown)


pont-us (approved):
> I propose to go for a new configuration as soon as xcube runs on python 3.10

I agree. Amongst other things, we should consider using micromamba rather than mamba installed via miniconda, which would save us about three minutes of set-up time.

Anyway, tests are clearly passing here, so approved :)

forman (approved):

PR #786, opened 2023-01-30
Fix affine_transform_dataset()

forman:
Function `xcube.core.resampling.affine.affine_transform_dataset()` now assumes that geographic coordinate systems are equal by default and hence a resampling based on an affine transformation can be performed.

Checklist:

* [x] AppVeyor CI passes

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,15 +13,19 @@
   2. `includes`: if not given or if any pattern matches the identifier, 
      the identifier is reported.
 
+### Fixes
+
 * xcube CLI tools no longer emit warnings when trying to import
   installed packages named `xcube_*` as xcube plugins.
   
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -13,15 +13,19 @@
...
+* The `xcube.util.timeindex` module can now handle 0-dimensional 
+  `ndarray`s as indexers. This effectively avoids the warning 
+  `Can't determine indexer timezone; leaving it unmodified.`
   which was emitted in such cases.
 
-### Fixes
-
+* Function `xcube.core.resampling.affine.affine_transform_dataset()`
+  now assumes that geographic coordinate systems are equal by default and
+  hence a resampling based on an affine transformation can be performed.
```
gunbra32, line 28: do you say here that xcube assumes all geographic coordinate systems to be WGS84? If so, is this a good idea?

Issue #806, opened 2023-02-10
xcube server is not correctly computing self-referencing URLs


forman:
**Describe the bug**

xcube server is not correctly computing self-referencing URLs. 

**To Reproduce**

For example, when accessing 
<https://bc-api.dev.brockmann-consult.de/api/openapi.html> it produces the following error message because the computed URL
<http://bc-api.dev.brockmann-consult.de/api/openapi.html> (HTTP instead of HTTPS!) is wrong:

![image](https://user-images.githubusercontent.com/206773/218099541-08883e40-4ee4-402d-aee1-300ed1737509.png)

**Additional context**

Occurs in xcube 0.13.0

forman:
Fixed by #809

PR #809, opened 2023-02-13
Allow for absolute URL prefixes

forman:
We don't have a clear solution for #806 because we cannot know whether the actual protocol as seen by users is HTTP or HTTPS. We also cannot detect, whether the service URL is already a route configured by some proxy and hence the server's base URL would already contain a path component.

Therefore, the current solution in this PR is to allow for absolute URL prefixes. If a URL prefix is absolute, we ignore the protocol and host reported by the (Tornado) request and use it directly as base URL for constructing self-referencing URLs. 

Checklist:

* [x] AppVeyor CI passes

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,14 @@
 ## Changes in 0.13.1 (in development)
 
+### Fixes
+
+* The xcube server configuration parameters `url_prefix` and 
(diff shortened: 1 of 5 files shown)


forman:
FYI @pont-us @TejasMorbagal @TonioF

thomasstorm (approved):

AliceBalfanz (approved):
Looking forward to testing it :)

PR #829, opened 2023-03-08
`resample_in_space()` to encode grid mapping correctly

### LINKED ISSUE #822: resample_in_space does not align crs in result to target crs
**Describe the bug**
When using resample_in_space() to transform a source ds, here named SC, to a target_gm, the resulting grid mapping, in the below example named resampled_SC, is not identical to target_gm.

**To Reproduce**
image

**Expected behavior**
Expected behaviour is that the output of resample_in_space() has an identical grid mapping as target gm.
.

**Additional context**
DeepESDL

forman:
xcube's spatial resampling functions `resample_in_space()`, `affine_transform_dataset()`,  and `rectify_dataset()` exported from module `xcube.core.resampling` now encode the target grid mapping into the resampled datasets. (#822) 
  
This new default behaviour can be switched off by keyword argument `encode_cf=False`. The grid mapping name can be set by keyword argument `gm_name`. If `gm_name` is not given a grid mapping will not be encoded if all of the following conditions are true: 
  - The target CRS is geographic; 
  - The spatial dimension names are "lon" and "lat";
  - The spatial 1-D coordinate variables are named "lon" and "lat" and are evenly spaced. 

The encoding of the grid mapping is done according to CF conventions:
  - The CRS is encoded as attributes of a 0-D data variable named by `gm_name`
  - All spatial data variables receive an attribute `grid_mapping` that is set to the value of `gm_name`.

Closes #822

Checklist:

* [x] AppVeyor CI passes


pont-us (approved):

TonioF (approved):
Looks good. The one thing I did not like is that some of the notebooks (affine, coords, rectify_dataset) have no description cells. Even a single one would be nice.

PR #847, opened 2023-04-17
Order variables reported by xcube server

### LINKED ISSUE #835: `xcube server`: specify variable to be displayed by default in xcube viewer
When I publish a dataset via xcube server which is then visualised by xcube viewer, the variables are variables in the drop down menu are sorted alphabetically. And per default the first variable in the list is visualised. However, this is not always an variable of interest for the user of xcube viewer. 

I would like to be able to specify a variable which should be used per default in xcube viewer. To achieve this, I would like to be able to specify it in the xcube server configuration yaml file or also be able to indicate the default variable already in the datasets metadata. 

Currently there is no alternative to this. 

**Additional context**

forman:
It is now possible to enforce the order of variables reported by xcube server. The new server configuration key `Variables` can be added to dataset configurations. Is a list of wildcard patterns that determine the variables' order.

Closes #835

Checklist:

* [ ] AppVeyor CI passes

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,8 +5,9 @@
 
 * It is now possible to enforce the order of variables reported by 
   xcube server. The new server configuration key `Variables` can be added 
-  to dataset configurations. Is a list of wildcard patterns that 
-  determine the variables' order. (#835) 
+  to `Datasets` configurations. Is a list of wildcard patterns that 
+  determines the order of variables and the subset of variables to be 
+  reported. (#835) 
 
 ## Changes in 1.0.3
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,11 @@
 * Setting a dataset's `BoundingBox` in the server configuration 
   is now recognised when requesting the dataset details. (#845)
 
+* It is now possible to enforce the order of variables reported by 
+  xcube server. The new server configuration key `Variables` can be added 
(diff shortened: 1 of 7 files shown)


CHANGES.md
```diff
@@ -3,6 +3,11 @@
 * Setting a dataset's `BoundingBox` in the server configuration 
   is now recognised when requesting the dataset details. (#845)
 
+* It is now possible to enforce the order of variables reported by 
+  xcube server. The new server configuration key `Variables` can be added 
+  to dataset configurations. Is a list of wildcard patterns that 
+  determine the variables' order. (#835)
```
AliceBalfanz, line 9: Please add, that if for a dataset only a subset of variables is defined and **not** followed by `- "*"`, then only the specified variables are displayed.

AliceBalfanz (changes requested):
Please see comment. Furthermore, I am missing an update in the documentation. But this should not be blocking, I can create a PR for that.

CHANGES.md
forman: Good point!

forman:
> Please see comment. Furthermore, I am missing an update in the documentation. But this should not be blocking, I can create a PR for that.

I'll do that too.

AliceBalfanz (approved):
thank you for adding the documentation!

PR #855, opened 2023-05-25
Adapt GH workflow to use GH app

thomasstorm:
Adapted the GitHub workflow to use a GitHub app for authentication.

### DIFF
--- a/.github/workflows/xcube_workflow.yaml
+++ b/.github/workflows/xcube_workflow.yaml
@@ -99,9 +99,8 @@ jobs:
         uses: actions/checkout@v2
       - name: checkout-k8s
         run: |
-          mkdir k8s
-          cd k8s
           git clone https://x-access-token:${{ steps.get_installation_token.outputs.token }}@github.com/bc-org/k8s-configs.git
+          mv k8s-configs k8s
       - name: get-release-tag
         id: release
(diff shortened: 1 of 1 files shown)


.github/workflows/xcube_workflow.yaml
```diff
@@ -86,13 +86,21 @@ jobs:
...
+        id: get_installation_token
+        uses: tibdex/github-app-token@v1
         with:
+          app_id: ${{ secrets.TOKEN_PROVIDER_APP_ID }}
+          private_key: ${{ secrets.TOKEN_PROVIDER_KEY }}
           repository: bc-org/k8s-configs
-          token: ${{ secrets.API_TOKEN_GITHUB }}
-          path: k8s
+          # the installationId of the GitHub app we are using
+          installationId: 36950178
```
pont-us, line 97: I'd prefer `installationId` to be a secret too -- not sure whether it's actually sensitive, but seems to me like the kind of variable information that doesn't belong in version control. But I don't insist on it.

pont-us (approved):
Mainly reviewed by comparison with the already working <https://github.com/dcs4cop/xcube-geodb/pull/73>, only found one thing to complain about but even that one is optional, so approving :)

AliceBalfanz (approved):

Issue #856, opened 2023-06-02
conda install not working


gdkrmr:
```
[gkraemer@orange py]$ conda create -c conda-forge -n xcube xcube
Collecting package metadata (current_repodata.json): done
Solving environment: unsuccessful attempt using repodata from current_repodata.json, retrying with next repodata source.
Collecting package metadata (repodata.json): done
Solving environment: / 
```
just keeps on solving forever.

janneskruse:
Having the same issue. This happens is for any version returned via conda search -c conda-forge xcube. Also maybe interesting: On pip there is a package named [xcube 0.0.1](https://pypi.org/project/xcube/) with a different purpose, which can be confusing.

AliceBalfanz:
@gdkrmr and @janneskruse thanks for reporting, please consider using mamba instead of conda:
https://xcube.readthedocs.io/en/latest/installation.html#installation-using-mamba

we experience way faster installation with mamba :)

janneskruse:
> @gdkrmr and @janneskruse thanks for reporting, please consider using mamba instead of conda: https://xcube.readthedocs.io/en/latest/installation.html#installation-using-mamba
> 
> we experience way faster installation with mamba :)

Thank you @AliceBalfanz !
Mamba solved the xcube installation problem but now I am having some issues with _aiohttp_ throwing _AttributeError: module 'aiohttp' has no attribute 'http_exceptions'_

I am trying to open this cube:

```
from xcube.core.store import new_data_store
store = new_data_store("s3", root="deep-esdl-public", storage_options=dict(anon=True))
ds = store.open_data('ocean-1M-9km-1x1080x1080-1.4.0.zarr')

```

Do you know if I need a specific version of aiohttp?

AliceBalfanz:
@janneskruse with `aiohttp 3.8.4` there are no issues, which version do you use? 
By the way, I see you are using the deep-esdl cube, but you are not one of our deepesdl jupyterlab users yet - there you would have functioning python environments preconfigured already  :) Let us know, in case you are interested in being onboarded!

PR #862, opened 2023-06-06
Update installation instructions

### LINKED ISSUE #859: Update installation documentation
The installation instructions are rather outdated and still suggests conda (which is *extremely* slow -- see e.g. #856). Installation instructions should be updated to use current best practices (probably mamba via mamba-forge).

pont-us:
Closes #859.

Checklist:

* ~[ ] Add unit tests and/or doctests in docstrings~ n/a
* ~[ ] Add docstrings and API docs for any new/modified user-facing classes and functions~ n/a
* ~[ ] New/modified features documented in `docs/source/*`~ n/a
* [ ] AppVeyor CI passes


docs/source/installation.md
```diff
@@ -1,45 +1,77 @@
...
+also automatically install and manage xcube’s dependencies.
+You can find [installation instructions for mamba itself
+here](https://mamba.readthedocs.io/en/latest/installation.html), if you
+don’t already have it installed.
 
-## Installation from the conda package
+Note that, while it is technically possible to use the conda package manager
+as an alternative to mamba, we do not recommend this: xcube uses a large
+number of dependencies from the conda-forge channel, and it can take conda
+several hours to resolve these dependencies during installation.
```
forman, line 17: Please rephrase. This is too negative. 
I fear, this will distract people from using xcube. The list of xcube's dependencies is really not special (Pangeo!). It is just a pitty that the conda channel is so slow, and some dependencies are so large in size. We have also listed some dependency to pin versions, although we actually not import them on our own.

docs/source/installation.md
```diff
@@ -1,45 +1,77 @@
...
-using conda proves to be unacceptably slow, mamba can be used instead (see
-[Installation using mamba](#installation-using-mamba)).
+xcube releases are packaged as conda packages in the
+[conda-forge](https://conda-forge.org/) channel. It is recommended to install
+xcube into a conda environment using the
+[mamba package manager](https://github.com/mamba-org/mamba), which will
+also automatically install and manage xcube’s dependencies.
+You can find [installation instructions for mamba itself
+here](https://mamba.readthedocs.io/en/latest/installation.html), if you
+don’t already have it installed.
```
forman, line 12: Do not exclude conda. People may not want to change their package manager. Others already have a conda env and may want to put xcube on top. They cannot simply switch to mamba.

forman: Also mention micromamba (I only use micromamba) as an even slimmer version of mamba.

forman (changes requested):
Very good, but please see my comments.

AliceBalfanz (approved):

docs/source/installation.md
pont-us: @forman  I've rewritten to mention micromamba and to be less emphatic about conda's slowness. But I still think that it's important to mention that conda *is* slow, to avoid user complaints like #856 ("just keeps on solving forever" -- no, probably only for a few hours, but if a user's not expecting that then of course they won't wait that long).

docs/source/installation.md
pont-us: Done.

PR #871, opened 2023-07-27
Enable xcube server to allow for externally added place groups

thomasstorm:
After acceptance of this PR, xcube server will offer the function `add_place_group` in `places/context.py`, which allows plugins to add place groups from any sources.
See [xcube-geodb-places](https://github.com/dcs4cop/xcube-geodb-places) for such a plugin.


xcube/webapi/places/context.py
```diff
@@ -183,27 +191,49 @@ def _load_place_group(self,
...
+
+    @staticmethod
+    def check_sub_group_configs(place_group_config):
+        sub_place_group_configs = place_group_config.get("Places")
+        if sub_place_group_configs:
+            raise ApiError.InvalidServerConfig(
+                "Invalid 'Places' entry in a 'PlaceGroups' item:"
+                " not implemented yet"
+            )
+        # sub_place_group_configs = place_group_config.get("Places")
```
AliceBalfanz, line 219: can this be removed? I understand, it was there before, but still I dont understand

AliceBalfanz (approved):
Please see my comment. And I have adjusted the Notebook, and pushed directly to the branch. Would be great if you could review it @thomasstorm

xcube/webapi/places/context.py
thomasstorm: done

PR #880, opened 2023-08-17
Implement OGC API - Coverages

### LINKED ISSUE #879: Implement OGC API - Coverages
We need an implementation of OGC API - Coverages for OGC Testbed-19.

`xcube serve` should provide a standard-compliant OGC API - Coverages instance along with its other APIs.

 - Use the reference Python implementation in [pygeoapi](https://pygeoapi.io/) -- but this would be hard to integrate with xcube's existing server framework and pull in various additional dependencies.
 - Don't implement the API -- but this would largely exclude us from Testbed-19 participation, since OGC API - Coverages is a core component of the draft Geodatacube API.

**Additional context**
 - [Draft GDC API](https://m-mohr.github.io/geodatacube-api/)
 - [Testbed-19 call for participation(https://portal.ogc.org/files/?artifact_id=104098)

pont-us:
Closes #879 .

Checklist:

* [x] GitHub CI passes
* [x] AppVeyor CI passes


thomasstorm (commented):
I guess the reason you did not implement the endpoints listed in the screenshot below is that these endpoints are already served from the STAC API implementation we have someplace else, correct? In that case it's fine, otherwise I think they have to implemented as well.
image

xcube/webapi/ows/coverages/controllers.py
```diff
@@ -0,0 +1,312 @@
...
+    :return: a JSON representation of the coverage
+    """
+    return {
+        'id': collection_id,
+        'type': 'CoverageByDomainAndRange',
+        'envelope': get_collection_envelope(ctx, collection_id),
+        'domainSet': get_coverage_domainset(ctx, collection_id),
+        'rangeSet': {
+            'type': 'RangeSet',
+            'dataBlock': {'type': 'VDataBlock', 'values': ['TODO']},  # FIXME
```
TonioF, line 49: I think you should write more about why this is not added.

xcube/webapi/ows/coverages/routes.py
```diff
@@ -0,0 +1,202 @@
...
+    """
+    Describe the range type of a coverage
+
+    The range type describes the types of the data contained in the range
+    of this coverage. For a data cube, these would typically correspond to
+    the types of the variables or bands.
+    """
+    # noinspection PyPep8Naming
+    @api.operation(
+        operation_id='coveragesDomainSet',
```
TonioF, line 115:
```suggestion
        operation_id='coveragesRangeType',
```

xcube/webapi/ows/coverages/routes.py
```diff
@@ -0,0 +1,202 @@
...
+            get_collection_metadata(self.ctx.datasets_ctx, collectionId)
+        )
+
+
+@api.route('/catalog/collections/{collectionId}/coverage/rangeset')
+class CoveragesRangesetHandler(ApiHandler[CoveragesContext]):
+    """
+    Handle rangeset endpoint with a "not supported" response
+
+    This endpoint has been deprecated, but we handle it to make clear that
```
TonioF, line 148: Where does it say that it is deprecated? My only reference is https://m-mohr.github.io/geodatacube-api/#tag/OGC-API-Coverages/operation/describe-coverage-rangeset and I didn't see any mention about it being deprecated there.

thomasstorm: The standard document also says nothing about deprecation: https://docs.ogc.org/DRAFTS/19-087.html#coverage-paths

thomasstorm (changes requested):
Generally fine but please consider comments.

pont-us:
> I guess the reason you did not implement the endpoints listed in the screenshot below is that these endpoints are already served from the STAC API implementation we have someplace else, correct?

Correct, these are covered by STAC.

xcube/webapi/ows/coverages/routes.py
pont-us: Yes, this is brand-new, red-hot news which I first heard of at this week's Testbed-19 meeting :). There is a PR. planned to be merged today, for the GDC API draft at <https://github.com/m-mohr/geodatacube-api/pull/1> containing this change. The corresponding change in the Coverages draft is scheduled for a SWG vote some time soon (and apparently no opposition is expected).

The GDC PR also removes the metadata endpoint, but since I'd already implemented it when I learned of the revisions, and since it's potentially useful to have, I thought I may as well leave it in.

xcube/webapi/ows/coverages/routes.py
```diff
@@ -0,0 +1,202 @@
...
+            self.response.set_status(415)
+            return await self.response.finish(
+                f'Available media types: {", ".join(available_types)}\n'
+            )
+        elif content_type == 'text/html':
+            result = (
+                f'<html><title>Collection</title><body>'
+                f'<p>{collectionId}</p>'
+                f'</body></html>'
+            )
```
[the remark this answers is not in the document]
pont-us: Thanks, good catch. I've now updated the HTML response: it contains a preformatted block consisting of the entire JSON representation of the coverage, which must surely count as "equivalent" to the JSON as the spec requires ;-)

xcube/webapi/ows/coverages/controllers.py
pont-us: I agree, and I've now done so.

Issue #886, opened 2023-08-28
xcube Developer Guide describes outdated way for conda releases


thomasstorm:
The information in the xcube Developer Guide about preparing a conda release are outdated: While it is said that forks can be found under `https://github.com/dcs4cop/<repo-name>-feedstock`, it actually should say that conda-forge recommends cloning into a personal repository, i.e. under the maintainer's GitHub account.

pont-us:
Closed by #887

PR #887, opened 2023-08-28
fixing #886 (update developer guide)

thomasstorm:
See https://github.com/dcs4cop/xcube/issues/886

### DIFF
--- a/docs/source/devguide.md
+++ b/docs/source/devguide.md
@@ -343,6 +343,8 @@ Conda-forge packages are produced from a github feedstock repository belonging
 to the conda-forge organization. A repository's feedstock is usually located at 
 `https://github.com/conda-forge/<repo-name>-feedstock`, e.g., 
 `https://github.com/conda-forge/xcube-feedstock`.
+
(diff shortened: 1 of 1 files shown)


docs/source/devguide.md
```diff
@@ -342,20 +342,17 @@ These instructions are based on the documentation at
 Conda-forge packages are produced from a github feedstock repository belonging 
 to the conda-forge organization. A repository's feedstock is usually located at 
 `https://github.com/conda-forge/<repo-name>-feedstock`, e.g., 
-https://github.com/conda-forge/xcube-feedstock.
+`https://github.com/conda-forge/xcube-feedstock`.
```
pont-us, line 345:
```suggestion
`https://github.com/conda-forge/xcube-feedstock`.

Usually, a conda-forge bot will create a feedstock pull request automatically for each new GitHub release, and the maintainers only need to merge it (potentially after some manual improvements). The manual procedure below is only needed if there isn't time to wait for the bot, or if an additional conda-forge build has to be done without a new GitHub release.
```

pont-us (approved):
I suggested an addition to make it clear that our default method is to let the bot create the PR. Approving in any case :)

Issue #869, opened 2023-07-25
Fix "Invalid projection: CRS84" error with gdal >3.6.2


pont-us:
**Describe the bug**

The `gdal` package is currently pinned to <=3.6.2 (since commit 5c767f03b9bb) in environment.yml, because running the xcube unit test suite with gdal >=3.6.3 produces 106 test failures due to the following error:

```text
pyproj.exceptions.CRSError: Invalid projection: CRS84: (Internal Proj Error: proj_create: unrecognized format / unknown name)
```

See [this CI job output](https://github.com/dcs4cop/xcube/actions/runs/4894548230/job/13341638274) for examples of these errors.

We should make xcube work with more recent GDAL versions.

**To Reproduce**
Steps to reproduce the behaviour:
1. Check out the latest commit on the xcube `master` branch.
2. Edit `environment.yml` to remove the `<3.6.3` restriction from `gdal`
3. `mamba env create --name gdalbug  --file environment.yml`
4. `conda activate gdalbug`
5. `pytest`

**Expected behaviour**
All tests should pass.

**Additional context**
After some investigation, I discovered that the error is triggered by a change in the behaviour of the `pyproj.crs.CRS.from_string` function.

- Up to GDAL 3.6.2,`pyproj.crs.CRS.from_string("CRS84")` returns the expected CRS.
- From GDAL 3.6.3 to 3.7.0 (current latest release), `pyproj.crs.CRS.from_string("CRS84")` raises the `CRSError` described above.

I haven't found any documentation or explanation of this change. Fortunately there's a simple fix: with newer GDAL versions, the following still works:

```python
pyproj.crs.CRS.from_string("OGC:CRS84")
```

Note the "OGC:" authority prefix in the argument. The resulting CRS is identical with that produced from the "CRS84" argument with previous GDAL versions. The specifier "OGC:84" also produces the same CRS, and works with both older and newer GDAL versions.

I've done some quick experiments with the latest GDAL (3.7.0) in an xcube environment, changing instances of  `pyproj.crs.CRS.from_string("CRS84")` to add the "OGC:" prefix, and it seems to reliably fix the unit test failures.  

However, there are lots of instances of this invocation in the codebase, and a few tests and runtime checks need to be updated to accommodate the change, so a little time and care will be needed to implement and test the fix properly.

Partial list of files in which the "CRS84" specifier needs to be updated:

- xcube/core/gridmapping/base.py
- xcube/core/tilingscheme.py
[...]

PR #896, opened 2023-09-22
Gdal version issue solved

ruchimotwaniBC:
[Description of PR]
Solves #869, the issue of gdal version.

Checklist:

* [x] GitHub CI passes
* [x] AppVeyor CI passes

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,7 @@
 * Added a basic implementation of the draft version of OGC API - Coverages
 
 ### Fixes
-* Resolved the issue for CRS84 error due to latest version of gdal
+* Resolved the issue for CRS84 error due to latest version of gdal (#869)
 
 ### Other changes
 * `update_dataset_attrs` can now also handle datasets with CRS other than 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 * Added a basic implementation of the draft version of OGC API - Coverages
 
 ### Fixes
+* Resolved the issue for CRS84 error due to latest version of gdal
 
 ### Other changes
 * `update_dataset_attrs` can now also handle datasets with CRS other than 
--- a/environment.yml
+++ b/environment.yml
@@ -17,7 +17,7 @@ dependencies:
   - fiona >=1.8
   - fontconfig
   - fsspec >=2021.6
-  - gdal >=3.0,<3.6.3
+  - gdal >=3.0
   - geopandas >=0.8
   - jdcal >=1.4
   - jsonschema >=3.2
--- a/environment.yml
+++ b/environment.yml
@@ -44,7 +44,7 @@ dependencies:
   - tornado >=6.0
   - urllib3 >=1.26
   - werkzeug <2.2  # >=2.2 slows down S3 tests (deps: moto->flask->werkzeug)
-  - xarray >=2022.6
+  - xarray >=2022.6,<2023.09.0
   - zarr >=2.11
   # Required by Coiled
   # These are very likely transitive deps anyway
--- a/environment.yml
+++ b/environment.yml
@@ -44,7 +44,7 @@ dependencies:
   - tornado >=6.0
   - urllib3 >=1.26
   - werkzeug <2.2  # >=2.2 slows down S3 tests (deps: moto->flask->werkzeug)
-  - xarray >=2022.6,<2023.09.0
+  - xarray >=2022.6,<2023.09.0 # because of issue https://github.com/dcs4cop/xcube/issues/897
   - zarr >=2.11
   # Required by Coiled
   # These are very likely transitive deps anyway
--- a/test/core/mldataset/test_base.py
+++ b/test/core/mldataset/test_base.py
@@ -5,6 +5,8 @@ from xcube.core.mldataset import BaseMultiLevelDataset
 from xcube.core.tilingscheme import TilingScheme
 from .helpers import get_test_dataset
 
+_CRS84 = 'OGC:CRS84'
+
 
 class BaseMultiLevelDatasetTest(unittest.TestCase):
     def test_basic_props(self):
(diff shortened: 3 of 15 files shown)


test/core/store/fs/test_registry.py
```diff
@@ -341,7 +342,7 @@ def _assert_dataset_supported(
             requested_dtype_alias: Optional[str],
             expected_dtype_aliases: Set[str],
             expected_return_type: Union[Type[xr.Dataset],
-                                        Type[MultiLevelDataset]],
+            Type[MultiLevelDataset]],
```
AliceBalfanz, line 345: I believe the prior existing indent was correct.

xcube/core/timeseries.py
```diff
@@ -59,7 +59,7 @@
     AGG_COUNT: CAN_COMPUTE
 }
 
-_CRS84 = pyproj.CRS.from_string('CRS84')
+_CRS84 = pyproj.CRS.from_string("OGC:CRS84")
```
AliceBalfanz, line 62: We usually use single quotes for consitency, as far as possible. However, as you can tell from older code - we are mixing double quotes and single quotes... But would be great not to change existing single quotes into double quotes :)

test/core/store/fs/test_registry.py
ruchimotwaniBC: Yes, true and I didnt notice this change. I do it :)

test/core/mldataset/test_base.py
```diff
@@ -5,6 +5,8 @@
 from xcube.core.tilingscheme import TilingScheme
 from .helpers import get_test_dataset
 
+_CRS84 = 'OGC:CRS84'
```
AliceBalfanz, line 8: @pont-us do we have a convention for naming constants? I feel like it is the first time I see a "_" prior to a constant name, but I might be wrong.

CHANGES.md
```diff
@@ -5,6 +5,7 @@
 * Added a basic implementation of the draft version of OGC API - Coverages
 
 ### Fixes
+* Resolved the issue for CRS84 error due to latest version of gdal
```
AliceBalfanz, line 8:
```suggestion
* Resolved the issue for CRS84 error due to latest version of gdal (#869)
```

test/core/mldataset/test_base.py
TonioF: The leading underscore denotes constants that are only used within the module in which they are declared. This is used in some places in xcube and the plugins. I did not see any rule in our conventions that en- or discourages its use. Personally, I like it as it gives some order.

TonioF (commented):
More of a comment than a full review: We have a module constants.py. It would be best to declare the CRS85 string there once and then have every place that requires it import it. If there ever would be another change, we would then just have to change the one place.

test/core/mldataset/test_base.py
pont-us: The leading underscore is recommended in PEP-8, so implicitly part of our conventions :)

pont-us:
> More of a comment than a full review: We have a module constants.py. It would be best to declare the CRS85 string there once and then have every place that requires it import it. If there ever would be another change, we would then just have to change the one place.

In principle I agree that this would have been preferable, but I don't think that it's worth the additional effort to optimize it now. This PR already introduces module-level constants, which is at least a start. In the unlikely event that the identifier changes again, we can optimize to a global constant at that point.

xcube/core/tilingscheme.py
```diff
@@ -36,11 +36,11 @@
     'EPSG:3857',
     'urn:ogc:def:crs:OGC:1.3:EPSG::3857',
 )
-GEOGRAPHIC_CRS_NAME = 'CRS84'
+GEOGRAPHIC_CRS_NAME = 'OGC:CRS84'
 GEOGRAPHIC_CRS_ALIASES = (
     'EPSG:4326',
     'urn:ogc:def:crs:OGC:1.3:EPSG::4326',
-    'CRS84',
+    'OGC:CRS84',
```
pont-us, line 43:
```suggestion
    'OGC:CRS84',
    'CRS84',
```

This is a list of aliases recognized by the class, so we should keep the old identifier as well as adding the new one.

pont-us (changes requested):
One change requested in code comments.

You should also update the xarray line in `environment.yml` to

```yaml
  - xarray >=2022.6,<2023.09.0
```

-- See #897 . This is of course unrelated to your changes but will make ensure that the tests run green.

I still think that the GitHub unit test failures are random and transient, so let's see if they go away on the next re-run. They do need to be fixed properly at some point, but not in this PR...

environment.yml
```diff
@@ -44,7 +44,7 @@ dependencies:
   - tornado >=6.0
   - urllib3 >=1.26
   - werkzeug <2.2  # >=2.2 slows down S3 tests (deps: moto->flask->werkzeug)
-  - xarray >=2022.6
+  - xarray >=2022.6,<2023.09.0
```
AliceBalfanz, line 47: why is restricting xarray needed?

test/core/test_normalize.py
```diff
@@ -21,6 +21,7 @@
 from xcube.core.normalize import normalize_coord_vars
 from xcube.core.normalize import normalize_dataset
 from xcube.core.normalize import normalize_missing_time
+from xcube.constants import CRS_CRS84
```
AliceBalfanz, line 24: typo? shouldn't it be CRS84?

xcube/constants.py
```diff
@@ -64,6 +65,9 @@
     AUTHORITY["EPSG","4326"]]
 """
 
+CRS84 = 'OGC:CRS84'
+CRS_CRS84 = pyproj.crs.CRS.from_string('OGC:CRS84')
```
pont-us, line 69:
```suggestion
CRS_CRS84 = pyproj.crs.CRS.from_string(CRS84)
```

I'm being pedantic, but this is a slight improvement :)

pont-us (approved):
Looks good! I have one very minor suggestion for improvement, but I don't need to review again, so approving.

xcube/constants.py
ruchimotwaniBC: Ya but make sense, can do it :)

environment.yml
AliceBalfanz, line 47:
```suggestion
  - xarray >=2022.6,<2023.09.0 # because of issue https://github.com/dcs4cop/xcube/issues/897
```

AliceBalfanz (approved):
Approving, but please add a comment to the xarray restriction which explains why the versions are restricted :)

TejasMorbagal (approved):

PR #907, opened 2023-10-24
new workflow to build and push docker image

TejasMorbagal:
[Description of PR]

This PR introduces a new workflow which builds and pushes a new xcube docker image. This is a workaround for the timeout errors in unit test which blocks us from making a release.

Checklist:

~* [ ] Add unit tests and/or doctests in docstrings~
~* [ ] Add docstrings and API docs for any new/modified user-facing classes and functions~
~* [ ] New/modified features documented in `docs/source/*`~
~* [ ] GitHub CI passes~
~* [ ] AppVeyor CI passes~
~* [ ] Test coverage remains or increases (target 100%)~

### DIFF
--- a/.github/workflows/xcube_build_docker.yaml
+++ b/.github/workflows/xcube_build_docker.yaml
@@ -1,0 +1,37 @@
+name: xcube docker build
+
+on:
+  workflow_dispatch:
+
+env:
+  APP_NAME: xcube
+  ORG_NAME: bcdev
+  IMG_REG_NAME: quay.io
+
+jobs:
+  build-docker-image-and-push:
+    runs-on: ubuntu-latest
+    # Build the docker image and push to quay.io
+    name: build-docker-image
+    steps:
+      - name: git-checkout
+        uses: actions/checkout@v2
+      # Determine release tag from git ref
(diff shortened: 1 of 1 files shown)


thomasstorm (approved):
Looks good!

TonioF (commented):
The script itself looks good, but what about that update-version part? Shouldn't that also be executed when a docker image is built?

TejasMorbagal:
> The script itself looks good, but what about that update-version part? Shouldn't that also be executed when a docker image is built?

My thoughts were that the ArgoCD image updater automatically updates the .argo files which is then used by deployment to fetch the new docker image. We no longer use docker image tags or hash directly. Do you agree on this?

TonioF:
> > The script itself looks good, but what about that update-version part? Shouldn't that also be executed when a docker image is built?
> 
> My thoughts were that the ArgoCD image updater automatically updates the .argo files which is then used by deployment to fetch the new docker image. We no longer use docker image tags or hash directly. Do you agree on this?

If we don't need this, it's fine. But shouldn't the update section be removed from the other workflow then, too?

pont-us (approved):

PR #912, opened 2023-10-26
build docker image from latest release tag

TejasMorbagal:
[Description of PR]

- This PR updates the workflow which builds docker image using the latest stable release tag. The plugins are not installed by intention to address missing packages such as rasterio,geopandas which gets uninstalled during installation of plugins.

Checklist:
~* [ ] Add unit tests and/or doctests in docstrings~
~* [ ] Add docstrings and API docs for any new/modified user-facing classes and functions~
~* [ ] New/modified features documented in `docs/source/*`~
~* [ ] Changes documented in `CHANGES.md`~
* [ ] GitHub CI passes
* [ ] AppVeyor CI passes


TejasMorbagal:
> I am quite certain the reason why some packages are missing is because some of the plugin environments still try to limit gdal. In the dockerfile we install the environment dependencies and then install xcube and the plugins from source (python setup.py install or pipi install .), meaning that conda does not know about restrictions imposed by xcube or any of the previously installed plugins. Given that we (a) are only dealing with released versions and (b) the script is run manually, I think the dockerfile could be changed to install xcube and its plugins from conda-forge rather than source.

We won't be able to have a docker image for dev releases with this approach, and the manual triggered workflow is a short-term workaround till we fix the failing unit tests.

TonioF:
> > I am quite certain the reason why some packages are missing is because some of the plugin environments still try to limit gdal. In the dockerfile we install the environment dependencies and then install xcube and the plugins from source (python setup.py install or pipi install .), meaning that conda does not know about restrictions imposed by xcube or any of the previously installed plugins. Given that we (a) are only dealing with released versions and (b) the script is run manually, I think the dockerfile could be changed to install xcube and its plugins from conda-forge rather than source.
> 
> We won't be able to have a docker image for dev releases with this approach, and the manual triggered workflow is a short-term workaround till we fix the failing unit tests.

Fair enough. It is still very unfortunate that by installing a plugin we can override dependency restrictions. We now need to change the environments in some of the plugins, just as we had to change them when the gdal issue occurred. This shouldn't be necessary.

TejasMorbagal:
> > > I am quite certain the reason why some packages are missing is because some of the plugin environments still try to limit gdal. In the dockerfile we install the environment dependencies and then install xcube and the plugins from source (python setup.py install or pipi install .), meaning that conda does not know about restrictions imposed by xcube or any of the previously installed plugins. Given that we (a) are only dealing with released versions and (b) the script is run manually, I think the dockerfile could be changed to install xcube and its plugins from conda-forge rather than source.
> > 
> > 
> > We won't be able to have a docker image for dev releases with this approach, and the manual triggered workflow is a short-term workaround till we fix the failing unit tests.
> 
> Fair enough. It is still very unfortunate that by installing a plugin we can override dependency restrictions. We now need to change the environments in some of the plugins, just as we had to change them when the gdal issue occurred. This shouldn't be necessary.

Yes, that's right. The idea of not installing plugins was also motivated by the fact that none of our viewer deployments actually needs plugins installed. So, why do we install plugins in the docker image ? This also makes the docker image larger.

pont-us (approved):

thomasstorm (approved):

AliceBalfanz (approved):

ruchimotwaniBC (approved):

PR #918, opened 2024-01-16
fixing GitHub CI #899

### LINKED ISSUE #899: Intermittent MaxRetryError (and others) in CI unit test runs
**Describe the bug**

Recently, CI unit test suite jobs have been producing increasingly frequent intermittent test failures due to time-outs and excessive retries. The problem seems to occur most frequently on GitHub, but occasionally also on AppVeyor. This issue was prompted by [this GitHub actions run](https://github.com/dcs4cop/xcube/actions/runs/6340782131/job/17222954166) which produced the following error:

```
FAILED test/webapi/ows/stac/test_routes.py::StacRoutesTest::test_fetch_catalog_collection_single_items - 
urllib3.exceptions.MaxRetryError: HTTPConnectionPool(host='localhost', port=34113): Max retries exceeded with url: 
/ogc/collections/demo/items (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f8a94db71d0>: 
Failed to establish a new connection: [Errno 111] Connection refused'))
```

We can use this issue to collect further instances of the problem.

So far our fix for such errors has been "re-run the job and hope it goes away", which it generally does, but this is turning into something of a time sink.

**To Reproduce**
Steps to reproduce the behavior: keep re-running the GitHub unittest job until an error occurs.

**Expected behavior**
All tests should pass reliably on every CI run.

**Additional context**
With luck, this might be fairly easily fixable by tweaking some back-off / time-out / retry parameters in `ServerTestCase` or similar.

thomasstorm:
This PR fixes #899. The reason for the test failures in GitHub has been a race condition: the server starts in the background, and on another thread, the tests use it. The fix ensured that the server has started before the setUp() method terminates.

The checklist items below do not apply mostly, as there are no new features to be documented.

Checklist:

* [x] GitHub CI passes
* [ ] AppVeyor CI passes

### DIFF
--- a/xcube/server/testing.py
+++ b/xcube/server/testing.py
@@ -31,6 +31,7 @@ from typing import Dict, Optional, Any, Union, Tuple
 
 import urllib3
 from tornado.platform.asyncio import AnyThreadEventLoopPolicy
+from urllib3.exceptions import MaxRetryError
 
 from xcube.server.server import Server
 from xcube.server.webservers.tornado import TornadoFramework
@@ -66,13 +67,20 @@ class ServerTestCase(unittest.TestCase, ABC):
 
         self.http = urllib3.PoolManager()
 
+        # Taking care that server is fully started until tests make requests.
+        # Fixing https://github.com/dcs4cop/xcube/issues/899
+        elapsed_time = 0.0
         while True:
-            # noinspection PyBroadException
+            if elapsed_time > 60:
+                self.fail('Server did not respond after one minute. '
+                          'Failing the test.')
             try:
                 self.fetch('/')
                 break
-            except Exception:
-                time.sleep(0.1)
+            except (MaxRetryError, TimeoutError):
+                sleep_time = 0.1
+                time.sleep(sleep_time)
+                elapsed_time += sleep_time
 
     def tearDown(self) -> None:
         self.server.stop()
--- a/xcube/server/testing.py
+++ b/xcube/server/testing.py
@@ -23,6 +23,7 @@ import asyncio
 import json
 import socket
 import threading
+import time
 import unittest
 from abc import ABC
 from contextlib import closing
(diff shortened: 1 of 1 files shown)


xcube/server/testing.py
```diff
@@ -65,6 +66,14 @@ def setUp(self) -> None:
 
         self.http = urllib3.PoolManager()
 
+        while True:
+            # noinspection PyBroadException
+            try:
+                self.fetch('/')
+                break
+            except Exception:
```
forman, line 74: This will loop forever if for some (other) reasons `fetch` raises.
Why this broad exception type? Isn't there a more concrete type when running into the race condition?

xcube/server/testing.py
```diff
@@ -65,6 +66,14 @@ def setUp(self) -> None:
...
+                time.sleep(0.1)
```
forman, line 75:
```suggestion
                # Intermittent MaxRetryError (and others) in CI unit test runs
                # https://github.com/dcs4cop/xcube/issues/899
                time.sleep(0.1)
```

xcube/server/testing.py
forman: E.g.  `MaxRetryError` and others.

forman (changes requested):

xcube/server/testing.py
thomasstorm: Thanks, adressed this.

xcube/server/testing.py
thomasstorm: Added reference to the issue before the respective code block.

pont-us:
I edited the title, since I assume "#886" was a typo for "#899".

forman (approved):

xcube/server/testing.py
```diff
@@ -65,6 +67,15 @@ def setUp(self) -> None:
 
         self.http = urllib3.PoolManager()
 
+        # Taking care that server is fully started until tests make requests.
```
pont-us, line 70:
```suggestion
        # Taking care that server is fully started before tests make requests.
```

TonioF (approved):
I see this rather as a workaround than a proper fix. In line 59, we instantiate an AnyThreadEventLoopPolicy which has been deprecated (see https://www.tornadoweb.org/en/stable/asyncio.html#tornado.platform.asyncio.AnyThreadEventLoopPolicy). I assume we'd need to remove this and that we still will encounter issues in the future. I approve this PR as it solves our momentary problems, but I suggest not to close the issue.

PR #940, opened 2024-03-05
Added tileLevelMin/Max and colormap details to STAC xcube:data_vars

edd3x:
[Description of PR]
Currently, our web visualisation development team have been using the `/datasets` endpoint from the xcube server API which provides `tileLevelMin`, `tileLevelMax`, `colorBarName`, `colorBarMax` and `colorBarMin` per data variable as part of the JSON output. We are however changing to `ows.stac` via the` /ogc` endpoint because it's faster, but the STAC JSON output does not provide the above-listed attributes.

What I have done here is to use the `DatasetsContext` to derive the tiling scheme and get the colormap details using `derive_tiling_scheme()` and `get_color_mapping()` in `xcube/webapi/ows/stac/controllers.py` so the `xcube:data_vars` in the STAC item JSON output will return the additional variables like `tileLevelMin`, `tileLevelMax`, `colorBarName`, `colorBarMax` and `colorBarMin`.

To keep the structure of `xcube:coords` in the cube properties, I renamed the method for generating it to `_get_xc_coordinates` and kept the `_get_xc_variables` for the `xcube:data_vars`.

Checklist:

* [x] New/modified features documented in `xcube/webapi/ows/stac/controllers.py`
* [ ] GitHub CI passes
* [ ] AppVeyor CI passes

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,28 @@
 ## Changes in 1.4.1 (in development)
 
 ### Enhancements
+* Improved xcube Server's STAC API:
+  * Added the `tileLevelMin`, `tileLevelMax`, `colorBarName`, `colorBarMax` and `colorBarMin` attributes to the `xcube:data_vars` property returned in a STAC item JSON schema of a dataset's data
+  variables.
+  * The `xcube:data_vars` in a STAC item will display the following example:
+   
+   ```json
+    "xcube:data_vars": [
+                    {
+                    "name": "gpp",
+                    "dtype": "float32",
+                    "dims": [],
+                    "chunks": [],
+                    "shape": [],
+                    "attrs": {},
+                    "tileLevelMax": 11, 
+                    "tileLevelMin": 10,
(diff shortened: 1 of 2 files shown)


forman (commented):
Hi @edd3x, thanks for the PR. The extension to data variables is very useful and code looks good!
In fact, we plan to deprecate the `/datasets` endpoint and use only STAC. Then, I'm sure we'll need to enrich the data variable object by even more information. To be compatible with such an extension, we might put the extra information into a dedicated object, either  as an object in `attrs` or as new own property `extra`  or similar. We'll let you know here asap. Stay tuned and thanks!

forman:
Hey @edd3x, FYI, we haven't started the activity yet, but it is still on the roadmap for spring 2024.

edd3x:
> Hey @edd3x, FYI, we haven't started the activity yet, but it is still on the roadmap for spring 2024.


Thanks @forman for the update, looking forward to the release

PR #947, opened 2024-03-15
Fix dependency discrepancies in environment.yml

### LINKED ISSUE #945: Resolve/document dependency discrepancies
conda-forge releases (e.g. [here](https://github.com/conda-forge/xcube-feedstock/pull/48) ) trigger a dependency analysis which always finds a significant number of discrepancies between the dependencies listed in `meta.yaml`. Automated dependency anaylsis isn't 100% reliable, but we should resolve as many of these as possible and document the rest so we don't waste time looking into them again. for the 1.4.1 release, the list is as follows:

### Packages found by source code inspection but not in the meta.yaml:
- cftime
- numcodecs
- botocore

### Packages found in the meta.yaml but not found by source code inspection:
- werkzeug
- scipy
- adlfs
- python-blosc
- fontconfig
- lz4
- requests-oauthlib

| Package | Comments | Action |
| --- | --- | --- |
| `cftime` | imported in select.py etc. | add to environment.yml and meta.yaml |
| `numcodecs` | imported in zarrstore/generic.py | add to environment.yml and meta.yaml |
| `botocore` | imported in dsio.py | add to environment.yml and meta.yaml |
| `werkzeug` | only imported in tests | remove from meta.yaml |
| `scipy` | not imported | remove from environment.yml and meta.yaml |
| `adlfs` | not imported, but has comment ‘for azure blob filesystem’ | clarify |
| `python-blosc` | not imported, but has comment ‘required by Coiled’ | clarify -- see below |
| `fontconfig` | not imported | remove from environment.yml and meta.yaml |
| `lz4` | not imported, but has comment ‘required by Coiled’ | clarify -- see below |
| `requests-oauthlib` | not imported | remove from environment.yml and meta.yaml |

Notes:

- We'll also need to update `DEFAULT_DEPENDENCY_NAMES` in `versions.py`.
- From the above list, it appears that we specify explicit dependencies on two transitive (soft?) dependencies of the `coiled` package, but the `coiled` package is only a soft dependency (i.e. we try to import locally at runtime if/when it's actually needed and raise an error if it's not available). If these are hard dependencies of coiled, we can probably omit them entirely: if coiled is importable, they must be present. If they're soft dependencies of coiled which we require for our use case, we should handle them the same way we handle the `coiled` package itself -- i.e. check at runtime if/when required.

pont-us:
Fixes #945.

Checklist:

* ~[ ] Add docstrings and API docs for any new/modified user-facing classes and functions~ n/a
* [x] GitHub CI passes
* [x] AppVeyor CI passes

### DIFF
--- a/xcube/core/store/fs/accessor.py
+++ b/xcube/core/store/fs/accessor.py
@@ -132,9 +132,9 @@ class FsAccessor:
                 root,
                 params
             )
-        except (ValueError, ImportError):
-            raise DataStoreError(f"Cannot instantiate"
-                                 f" filesystem {protocol!r}")
+        except (ValueError, ImportError) as error:
+            raise DataStoreError(
+                f"Cannot instantiate filesystem {protocol!r}: {error}")
 
     @classmethod
     def add_storage_options_to_params_schema(
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,8 @@
 ## Changes in 1.4.2 (in development)
 
+* Update dependencies to better match imports; remove the defaults channel;
+  turn adlfs into a soft dependency. (#945)
+
 ## Changes in 1.4.1
 
 ### Enhancements
--- a/docs/source/installation.md
+++ b/docs/source/installation.md
@@ -144,3 +144,16 @@ docker ps
 This will start a service in the background which can be accessed 
 through port 8001, as the startup of a service is configured as default
 behaviour.
+
+## Installing soft dependencies
+
+In addition to xcube’s ‘hard’ dependencies, which the standard installation
+methods install automatically when required, there are ‘soft’ dependencies.
+These are packages which are not required to install or use xcube, but enable
+additional functionality if they are present. Users who need this functionality
+should install the requisite soft dependencies explicitly. xcube’s current soft
+dependencies are listed below.
+
+- `adlfs`: required by the abfs data store, which is used for access to
+  Azure Blob storage. Trying to create an abfs data store without `adlfs`
+  installed will raise an exception advising that you install it.
--- a/environment.yml
+++ b/environment.yml
@@ -41,7 +41,6 @@ dependencies:
   - shapely >=1.6
   - tornado >=6.0
   - urllib3 >=1.26
-  - werkzeug
   - xarray >=2022.6
   - zarr >=2.11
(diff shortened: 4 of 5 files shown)


xcube/core/store/fs/accessor.py
```diff
@@ -132,9 +132,9 @@ def load_fs(cls, params: Dict[str, Any]) \
                 root,
                 params
             )
-        except (ValueError, ImportError):
-            raise DataStoreError(f"Cannot instantiate"
-                                 f" filesystem {protocol!r}")
+        except (ValueError, ImportError) as error:
+            raise DataStoreError(
+                f"Cannot instantiate filesystem {protocol!r}: {error}")
```
forman, line 137: It would be nice, if we say that the reason for this is most likely a missing package that provides the `fsspec` implementation.

```suggestion
                f"Cannot instantiate filesystem {protocol!r}"
                " (most likely your environment is missing the"
                " required `fsspec` implementation): {error}"
            )
```

forman (commented):
Looks good, thanks! Please recognize my optional suggestion.

TonioF (changes requested):
If werkzeug is only used by test, I suggest moving it to the test section in the environment.yaml for clarity. Otherwise good!

TonioF (approved):

xcube/core/store/fs/accessor.py
pont-us: Fortunately, the error message that we interpolate from `fsspec` already supplies an informative and constructive error message if there’s a missing package. For example, in the case of `abfs`, the user will see:

> xcube.core.store.error.DataStoreError: Cannot instantiate filesystem 'abfs': Install adlfs to access Azure Datalake Gen2 and Azure Blob Storage

See [here](https://github.com/fsspec/filesystem_spec/blob/ecc576506567b8413a4c5402cbb06b28f1ba27bb/fsspec/registry.py#L62) for the error messages for other protocols -- they’re all similarly helpful. With this in mind, I think that the additional explanation on our side is superfluous if the problem is a missing package, and potentially misleading if it isn’t.

pont-us:
> If werkzeug is only used by test, I suggest moving it to the test section in the environment.yaml for clarity. Otherwise good!

Done!

PR #954, opened 2024-03-28
Converted more docstrings from reST to Google style

forman:
Converted remaining docstrings from reST to Google style.


pont-us (changes requested):

forman:
Closing in favor of #960

PR #966, opened 2024-04-10
Preparing v1.5.1 release

forman:


TejasMorbagal:
Doesn't work for me in my local, after pip install -e . and xcube serve. The viewer endpoint doesn't return the viewer interface but only a blank page

forman:
> Doesn't work for me in my local, after pip install -e . and xcube serve. The viewer endpoint doesn't return the viewer interface but only a blank page

Could you provide some more details please? What is shown in the browser's developer tools? Anything interesting in the server logs? What OS, browser?

TejasMorbagal:
> > Doesn't work for me in my local, after pip install -e . and xcube serve. The viewer endpoint doesn't return the viewer interface but only a blank page
> 
> Could you provide some more details please? What is shown in the browser's developer tools? Anything interesting in the server logs? What OS, browser?

works now!

TejasMorbagal (approved):

Issue #975, opened 2024-05-06
User-defined color bars


forman:
**Is your feature request related to a problem? Please describe.**

This is the backend implementation for https://github.com/xcube-dev/xcube-viewer/issues/334.

**Describe the solution you'd like**

Custom color bars are still passed using query parameter `cmap`, but in the case of custom color bars it is a JSON object with the following format: `{"name": <name>, "colors": <colors>}`, where `<name>` is a unique name and `<colors>` is a list pairs `[[<v1>,<c1>], [<v2>,<c2>], [<v3>,<c3>], ...]` that map a sample value to a `"#RRGGBBAA"` color.

PR #977, opened 2024-05-13
Categorical color mapping

forman:
* xcube server's tile API can now handle user-defined colormaps from xcube 
  viewer. Custom color bars are still passed using query parameter `cmap` to 
  endpoint `/tiles/{datasetId}/{varName}/{z}/{y}/{x}`,
  but in the case of custom color bars it is a JSON-encoded object with the 
  following format: `{"name": <str>, "type": <str>, "colors": <list>}`. (#975) 
  The object properties are
  - `name`: a unique name.
  - `type`: optional type of control values.
  - `colors`: a list of pairs `[[<v1>,<c1>], [<v2>,<c2>], [<v3>,<c3>], ...]` 
    that map a control value to a hexadecimal color value using CSS format
    `"#RRGGBBAA"`. 
  
  The `type` values are
  - `"node"`: control points are nodes of a continuous color gradient.
  - `"bound"`: control points form bounds that map to a color, which means
     the last color is unused.
  - `"key"`: control points are keys (integers) that identify a color.

* xcube server's tile API now allows specifying the data normalisation step 
  before a color mapping is applied to the variable data to be visualized.
  This affects endpoint `/tiles/{datasetId}/{varName}/{z}/{y}/{x}` and the WMTS
  API. The possible normalisation values are 
  - `lin`: linear mapping of data values between `vmin` and `vmax` to range 0 to 1
    (uses `matplotlib.colors.Normalize(vmin, vmax)`).
  - `log`: logarithmic mapping of data values between `vmin` and `vmax` to range 0 to 1
    (uses `matplotlib.colors.LogNorm(vmin, vmax)`).
  - `cat`: categorical mapping of data values into to indexes into the color mapping.
    (uses `matplotlib.colors.BoundaryNorm(categories)`). This normalisation
    currently only works with user-defined colormaps of type
    `key` or `bound` (see above).
  
  The normalisation can be specified in three different ways (in order): 
  1. As query parameter `norm` passed to the tile endpoint. 
  2. Property `Norm` in the `Styles/ColorMapping` element in xcube server configuration.
  3. Data variable attribute `color_norm`.

The xcube viewer part of this PR is PR https://github.com/xcube-dev/xcube-viewer/pull/338.

Checklist:

* [ ] GitHub CI passes
* [ ] AppVeyor CI passes


pont-us (approved):
Various change requests / suggestions but only minor stuff, so approving since I don't need to see it again.

PR #984, opened 2024-05-21
Addressed future warnings regarding dataset.dims

### LINKED ISSUE #981: FutureWarnings: return type of `Dataset.dims`
### Arising of FutureWarnings regarding the return type of ```Dataset.dims```

When running unit tests, the following warning occurred:

```FutureWarning: The return type of `Dataset.dims` will be changed to return a set of dimension names in future, in order to be more consistent with `DataArray.dims`. To access a mapping from dimension names to lengths, please use `Dataset.sizes`.```

Solve Issue by replacing  ```Dataset.dims``` with  ```Dataset.sizes```.

### PR DESCRIPTION
**Pull request by clarasb**
Closes #981 

Checklist:

* [x] GitHub CI passes
* [x] AppVeyor CI passes

### CODE REVIEW

**Review by forman — APPROVED** [review]
Thanks, all good!

**Review by pont-us — APPROVED** [review]
Looks fine and works fine! Looks as though it's [just in time](https://github.com/pydata/xarray/pull/8980) too.

Just as a note: a few of the `dims` -> `sizes` substitutions here were not really necessary. For example in `iproc.py`:

```python
        if "time" in dataset.dims:
```

The impending xarray change won't remove `dims`, but will change the functionality from returning a dimension->size mapping to a set of dimensions. So tests with the `in` operator will continue to work.

However, there's also no harm in changing to `sizes`, since with dicts `in` checks membership of the set of keys, so the functionality in this case is identical. So this is just an incidental comment rather than a change request :).

forman:
> Just as a note: a few of the `dims` -> `sizes` substitutions here were not really necessary.

@pont-us please feel free to prepare a related PR for your requested changes. It would be much appreciated. This was however not in the scope of @clarasb's PR.

PR #987, opened 2024-05-23
Links added, which will be shown on the left side panel on the PyPI xcube-core webpage

konstntokas:
Links added to the left sidebar on [PyPI xcube-core](https://pypi.org/project/xcube-core/). Note that this will be available after the next release. As a example view [TestPyPI xcube-core](https://test.pypi.org/project/xcube/1.6.0.dev12/) 

Checklist:

* [x] GitHub CI passes
* [x] AppVeyor CI passes

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -62,6 +62,9 @@
   release. To install xcube via the `pip` tool use `pip install xcube-core`,  
   since the name "xcube" is already taken on PyPI by another software. (#982)
 
+* Add project urls and classifiers to `setup.py` which will be shown in the
+  left sidebar onthe [PyPI xcube-core](https://pypi.org/project/xcube-core/) webpage.
+
 ## Changes in 1.5.1
 
 * Embedded [xcube-viewer 1.1.1](https://github.com/xcube-dev/xcube-viewer/releases/tag/v1.1.1).
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -62,8 +62,8 @@
   release. To install xcube via the `pip` tool use `pip install xcube-core`,  
   since the name "xcube" is already taken on PyPI by another software. (#982)
 
-* Add project urls and classifiers to `setup.py` which will be shown in the
-  left sidebar onthe [PyPI xcube-core](https://pypi.org/project/xcube-core/) webpage.
+* Add project urls and classifiers to `setup.py`, which will be shown in the
+  left sidebar on the [PyPI xcube-core](https://pypi.org/project/xcube-core/) webpage.
 
 ## Changes in 1.5.1
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -62,7 +62,7 @@
   release. To install xcube via the `pip` tool use `pip install xcube-core`,  
   since the name "xcube" is already taken on PyPI by another software. (#982)
 
-* Add project urls and classifiers to `setup.py`, which will be shown in the
+* Added project URLs and classifiers to `setup.py`, which will be shown in the
   left sidebar on the [PyPI xcube-core](https://pypi.org/project/xcube-core/) webpage.
 
 ## Changes in 1.5.1
--- a/setup.py
+++ b/setup.py
@@ -96,4 +96,32 @@ setup(
         ],
     },
     install_requires=requirements,
+    # these links will be shown in the left panel on PyPI
+    project_urls={
+        "documentation": "https://xcube.readthedocs.io/en/latest/",
+        "source": "https://github.com/xcube-dev/xcube",
+        "download": "https://pypi.org/project/xcube-core/#files",
+        "tracker": "https://github.com/xcube-dev/xcube/issues",
+        "release notes": "https://github.com/xcube-dev/xcube/releases"
+    },
+    # these classifiers will be shown in the left panel on PyPI
+    # they to not interfer with pip install and are meta data for the user
+    classifiers=[
+        'Development Status :: 5 - Production/Stable',
+        'Intended Audience :: Science/Research',
+        'Intended Audience :: Developers',
+        'License :: OSI Approved :: MIT License',
+        'Programming Language :: Python :: 3',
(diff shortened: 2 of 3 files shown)


konstntokas:
Please check the other [classifiers available on PyPI](https://pypi.org/classifiers/) and add them accordingly if desired.

CHANGES.md
```diff
@@ -62,6 +62,9 @@
   release. To install xcube via the `pip` tool use `pip install xcube-core`,  
   since the name "xcube" is already taken on PyPI by another software. (#982)
 
+* Add project urls and classifiers to `setup.py`, which will be shown in the
```
forman, line 65:
```suggestion
* Added project URLs and classifiers to `setup.py`, which will be shown in the
```

setup.py
```diff
@@ -96,4 +96,32 @@
...
     },
     install_requires=requirements,
+    # these links will be shown in the left panel on PyPI
+    project_urls={
+        "documentation": "https://xcube.readthedocs.io/en/latest/",
+        "source": "https://github.com/xcube-dev/xcube",
+        "download": "https://pypi.org/project/xcube-core/#files",
+        "tracker": "https://github.com/xcube-dev/xcube/issues",
+        "release notes": "https://github.com/xcube-dev/xcube/releases"
+    },
```
forman, line 106: Missing changelog and repository here. Look, in zappend's `pyproject.toml` we have 
```
[project.urls]
Documentation = "https://bcdev.github.io/zappend/"
Issues = "https://github.com/bcdev/zappend/issues"
Changelog = "https://github.com/bcdev/zappend/blob/main/CHANGES.md"
Repository = "https://github.com/bcdev/zappend"
```

konstntokas: Repository appears as 'Homepage' on [zappend PyPI](https://pypi.org/project/zappend/). PyPI recognizes "source" as GitHub and tags it with the GitHub sign on the [xcube-core TestPyPI](https://test.pypi.org/project/xcube/1.6.0.dev12/), which I like more. I'd prefer "source", except you have strong objections. 

I can add Changelog. However, this is very similar to "release notes" except that the changes of the current development are missing. Should we add both? 

Note that I took [numpy PyPI](https://pypi.org/project/numpy/) as inspiration. However, I just noticed that there links are with capital letters. Probably when using `pyproject.toml`, it modifies the field names to capitalized words which is not done with `setup.py`. I will change that accordingly.

forman: > I'd prefer "source", except you have strong objections.

I agree.

> I can add Changelog. However, this is very similar to "release notes" except that the changes of the current development are missing. Should we add both?

Yes please.

> I will change that accordingly.

Yes, if that works.

setup.py
```diff
@@ -96,4 +96,32 @@
         ],
     },
     install_requires=requirements,
```
forman, line 98: @konstntokas our requirements are currently empty. This is because there where issues when installing using conda, because setup used pip to resolve dependecies after conda installed them. For some packages, pip found the PyPI version more appropriate and installed them on top of the conda version. Which broke our env in many cases. We should give it a try and see whether those issues still exist.

forman: FYI @pont-us  @TejasMorbagal

TejasMorbagal: Yes, worth a try!

setup.py
konstntokas: All changes are applied, which can be viewed here on [TestPyPI xcube-core](https://test.pypi.org/project/xcube/1.6.0.dev14/)

setup.py
konstntokas: @forman and @TejasMorbagal: 

I used [pipreqs](https://pypi.org/project/pipreqs/) to determine the requirements. I deleted a few.

I then copied the requirements from `environment.yml` to `setup.py` for pip dependencies. I created a new conda environment with `conda env  create -f environment.yml` and installed xcube with `pip install .`. It gets problems with matplotlib-base. It cannot find a matching distribution for matplotlib-base. So I changed it to matplotlib which works fine. Any reason why we want to use matplotlib-base?

pont-us: @konstntokas Don't assume a 1:1 correspondence between PyPI and conda packages, or between either of them and the actual Python package name. In maybe >90% of cases they'll all be the same, but there are plenty of exceptions. Regarding your edits to `environment.yml`,

- `affine` is used directly in `xcube.core.geom`, so we need to keep it.
- We don't directly use `openssl`, so we can drop that. (Also, Python itself depends on openssl, so it will always be there as a transitive dependency anyway.)
- The conda package `matplotlib` includes a dependency on the Qt libraries, which are large and not needed by xcube. `matplotlib-base` excludes them, so we should stick with that. See [here](https://conda-forge.org/docs/maintainer/knowledge_base/#matplotlib).

pont-us (changes requested):
- Don't remove affine, since `xcube.core.geom` needs it
- Revert `matplotlib` to `matplotlib-base`, since we don't want pyqt.

setup.py
konstntokas: Thank you @forman and @pont-us 
- I added again `affine`
- I swiched back to `matplotlib-base` in `environment.yml` and kept `matplotlib` in `setup.py`, since `matplotlib-base` does not exists for `pip`. When installing with `pip` after the conda env is created from `environment.yml`, `matplotlib` is not reinstalled but `matplotlib-base` is recognized as 'Requirement already satisfied'.  So I think this is what we want, even though it is odd to me how the recognition works.

pont-us: @konstntokas Yes, it's maybe a little unexpected but pip checks for already installed dependencies directly in the Python path rather than in its own package database. But it's actually a useful behaviour when installing on top of a conda environment or a system Python with non-pip-installed packages. Nevertheless I tend to add the `--no-deps` flag, just in case, if I know that all dependencies are already present. A potential missing package is easier to deal with than a twice-installed package :)

pont-us (approved):

forman (approved):

PR #1003, opened 2024-06-05
Address issues with `resample_in_space()`

### LINKED ISSUE #1001: Performance issues with `resample_in_space()`

The following issues are recurring when working with function `resample_in_space()` of module  `xcube.core.resampling`:
  1. The returned target dataset may have coordinates of dtype float64 while the source or the target grid mapping was float32. Hence the values are no longer exactly the same.
  2. In the case of up-sampling, `resample_in_space()` takes considerable time to execute although it should be entirely lazy as it is a dask implementation.

Somehow address/mitigate the issues above.

forman:
Closes #1001

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -53,6 +53,16 @@
   A new keyword argument `max_depth` defines the maximum subdirectory depths 
   used to search for datasets in case `roots` is given. It defaults to `1`.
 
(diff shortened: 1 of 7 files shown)


forman:
Merging this so I can create a dev release. There is still room for improvement, e.g., add missing tests, improve docstrings. Will be done in another PR.

PR #1037, opened 2024-07-09
Add `data_type` to `open_data` method in `DataStore` class

### LINKED ISSUE #1030: Data store framework enhancement: add the optional argument `data_type` to the `open_data` method
If the user wants to open a data resource as a multi-level dataset, the user needs to define the `opener_id="mldataset:<format>:<protocol>"`, which is not user friendly.  

 In most of the cases format and protocol will be defined by the data resource. Therefore, it would be user friendly to add the optional argument `data_type`, where the user can decide between `"mldataset"` and `"dataset"`. The `opener_id` is then built dynamically.

konstntokas:
Closes #1030

Checklist:

* [x] GitHub CI passes
* [x] AppVeyor CI passes

### DIFF
--- a/test/core/gen2/local/test_usercode.py
+++ b/test/core/gen2/local/test_usercode.py
@@ -104,7 +104,7 @@ class CubeUserCodeExecutorTest(unittest.TestCase):
             "\n"
             "Failed validating 'minimum'"
             " in schema['properties']['value']:\n"
-            "    {'minimum': 1, 'type': 'integer'}\n"
+            "    {'type': 'integer', 'minimum': 1}\n"
             "\n"
             "On instance['value']:\n"
             "    0",
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,9 @@
   `pyproject.toml.` (related to #992) 
 * Normalisation with `xcube.core.normalize.normalize_dataset` fails when chunk encoding 
   must be updated (#1033)
+* Added `data_type` to `open_data` method in `DataStore`, which determines the
+  return value of the data set. Note that `opener_id` includes the `data_type`
+  at its first position and will override the `date_type` argument. (#1030)
 
 ## Changes in 1.6.0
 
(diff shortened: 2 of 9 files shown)


forman:
FYI @TonioF (xcube-cci), @pont-us (xcube-cds), @TejasMorbagal (xcube-cmems)

konstntokas:
As suggested, I included in keyword argument `data_type` in `**open_params` now. Further I added testing to the `open_data` method in the `ReferenceDataStore` class. Please check if I understood correctly, that the `open_data` method in the `ReferenceDataStore` class only returns `xr.Dataset`. However most of these tests are skipped because kerchunk is not installed (see e.g. https://github.com/xcube-dev/xcube/blob/main/test/core/store/ref/test_store.py#L178). Should we maybe add kerchunk to the testing section in the `environment.yml` ?

test/core/gen2/local/test_usercode.py
```diff
@@ -104,7 +104,7 @@ def test_class_bad_params(self):
             "\n"
             "Failed validating 'minimum'"
             " in schema['properties']['value']:\n"
-            "    {'minimum': 1, 'type': 'integer'}\n"
+            "    {'type': 'integer', 'minimum': 1}\n"
```
konstntokas, line 107: This test did not fail on my local machine, but did fail in the CI. Adjusting it, fixed the CI. Still I do not understand this behavior.

forman:
> Please check if I understood correctly, that the `open_data` method in the `ReferenceDataStore` class only returns `xr.Dataset`. 

True.

> However most of these tests are skipped because kerchunk is not installed (see e.g. https://github.com/xcube-dev/xcube/blob/main/test/core/store/ref/test_store.py#L178). Should we maybe add kerchunk to the testing section in the `environment.yml` ?

Let's do that!

test/core/gen2/local/test_usercode.py
forman: I guess, what is formatted here is not an ordered dict, so the order of fields may change from build to build or even run to run.

forman (approved):

PR #1040, opened 2024-07-15
Cleanup in module `xcube.core.varexpr`

forman:
Cleanup in module `xcube.core.varexpr`

Checklist:

* [ ] ~Changes documented in `CHANGES.md`~
* [x] GitHub CI passes
* [x] AppVeyor CI passes

### DIFF
--- a/xcube/core/varexpr/context.py
+++ b/xcube/core/varexpr/context.py
@@ -2,7 +2,6 @@
 # Permissions are hereby granted under the terms of the MIT License:
 # https://opensource.org/licenses/MIT.
 
-import ast
 import inspect
 from typing import Any, Callable, Union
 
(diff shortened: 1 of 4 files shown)


xcube/core/varexpr/context.py
```diff
@@ -171,6 +154,21 @@ def evaluate(self, var_expr: str) -> xr.DataArray:
         return result
 
 
-class VarExprValidator(ast.NodeTransformer):
-    def visit_Lambda(self, node):
-        raise VarExprError("lambda expressions are not supported")
+def _format_callable(name: str, fn: Union[np.ufunc, Callable]):
```
konstntokas, line 157: What is the reason for putting this outside of the class?

konstntokas (approved):
Yes looks cleaner than before. Would you like to update CHANGES.md?

forman:
> Would you like to update CHANGES.md?

No, this code and the former has not been released yet.

xcube/core/varexpr/context.py
forman: It is was a private static method and now it is a private function, calling is shorter as it does not require context object.

PR #1056, opened 2024-08-06
Force xcube Server to open multi-level datasets

forman:
- Changed xcube Server impl. to force opening multi-level datasets.
- Added schema for "data_type" parameter fro all filesystem data stores

Checklist:

* [x] GitHub CI passes
* [x] AppVeyor CI passes

### DIFF
--- a/xcube/webapi/datasets/context.py
+++ b/xcube/webapi/datasets/context.py
@@ -592,26 +592,42 @@ class DatasetsContext(ResourcesContext):
             data_store_pool = self.get_data_store_pool()
             data_store = data_store_pool.get_store(store_instance_id)
             data_id = dataset_config.get("Path")
-            open_params = dataset_config.get("StoreOpenParams") or {}
-            # Inject chunk_cache_capacity into open parameters
+            open_params = dict(dataset_config.get("StoreOpenParams") or {})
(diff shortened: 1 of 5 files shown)


xcube/webapi/datasets/context.py
```diff
@@ -592,26 +592,42 @@ def _open_ml_dataset(self, dataset_config: DatasetConfig) -> MultiLevelDataset:
...
             if isinstance(dataset, MultiLevelDataset):
+                # Expected, nominal case.
                 ml_dataset: MultiLevelDataset = dataset
             else:
+                # Fallback. Usually results in poor tile computation performance.
                 ml_dataset = BaseMultiLevelDataset(dataset)
+
             ml_dataset.ds_id = ds_id
+
         else:
```
konstntokas, line 631: As far as I understand this is the case for datasets calculated on the fly (in memory). Do we need to take care, that this is done as mldatasets as well, or is this covered already?

konstntokas (approved):
Looks good. I tried to test it on the xcube server demo, but could not recognize any performance increase. Maybe the demo example is too small. 

Also, it will be kind of hard to test that `"mldataset"` is selected, because the output of the function `_open_ml_dataset` stays the same. Would be interested in how you'd do it.

PR #1080, opened 2024-10-07
Xcube server STAC API now returns a conformant collection time extent

### LINKED ISSUE #1077: xcube v1.7.0 OGC API returns STAC non-conformant collection time extent.
**Describe the bug**

OGC API of xcube server v1.7.0 returns a temporal extent with an invalid format against the STAC Collection spec https://api.stacspec.org/v1.0.0/collections/#tag/Collections/operation/getCollections -  the endpoint returns temporal extent as list of lists of lists of strings.

**To Reproduce**
Steps to reproduce the behaviour:
Run xcube server v1.7.0 with the following config:

```DataStores:
  - Identifier: deep-esdl
    StoreId: s3
    StoreParams:
      root: deep-esdl-public
      storage_options:
        anon: false
    Datasets:
      - Identifier: hydrology
        Path: hydrology-1D-0.009deg-1x1102x966-3.0.2.levels
        Style: default
        Variables:
          - "SM"
          - "*"
        AccessControl:
          # Due to this it will only appear for unauthorized clients
          IsSubstitute: true
```

**Expected behaviour**
It should be a list of lists of strings

**Screenshots**
image

**Additional context**

konstntokas:
Closes #1077 

Checklist:

* [x] GitHub CI passes
* [ ] AppVeyor CI passes


test/webapi/ows/stac/test_controllers.py
```diff
@@ -148,6 +148,16 @@
 }
 
 
+def assert_equal_deep_dict(self, dict1, dict2):
```
konstntokas, line 151: I add this function to test for nested dictionaries.

test/webapi/ows/stac/test_controllers.py
```diff
@@ -174,7 +184,7 @@ def test_get_datasets_collection_item(self):
             "demo-1w",
         )
         self.assertIsInstance(result, dict)
-        self.assertCountEqual(self.read_json("stac-datacubes-item.json"), result)
+        assert_equal_deep_dict(self, self.read_json("stac-datacubes-item.json"), result)
```
konstntokas, line 187: Previously, it was only tested that the keys in the first layer of the dictionaries are equal. Now I test the full content of the dictionaries.

test/webapi/ows/stac/test_controllers.py
forman: Ok. However, you should keep track of the path and output it as part of the error message.

xcube/webapi/ows/stac/controllers.py
```diff
@@ -632,7 +632,7 @@ def _get_single_dataset_collection(
                     {"cellsCount": gm.size[1], "resolution": gm.xy_res[1]},
                 ],
             },
-            "temporal": {"interval": [[time_interval]], "grid": get_time_grid(dataset)},
+            "temporal": {"interval": [time_interval], "grid": get_time_grid(dataset)},
```
forman, line 635: Aha, the actual fix.

xcube/util/jsonencoder.py
```diff
@@ -102,7 +102,7 @@ def _convert_default(obj: Any) -> Any:
             elif np.issubdtype(obj.dtype, np.floating):
                 return float(obj)
             elif np.issubdtype(obj.dtype, np.datetime64):
-                return np.datetime_as_string(obj, timezone="UTC", unit="s")
+                return str(np.datetime_as_string(obj, timezone="UTC", unit="s"))
```
forman, line 105: Please add a comment why this is needed.

test/webapi/ows/stac/test_controllers.py
forman: But actually it is strange why this is needed. `TestCase.assertEquals()` actually does a deep equality comparison.

forman (commented):
Approved with comments. Please have a look.

test/webapi/ows/stac/test_controllers.py
konstntokas: `TestCase.assertEqual()` also checks for the order of the keys in the dictionary, which gives me an error.

konstntokas: `TestCase.assertEqual()` also checks for the order of the keys in the dictionary, which gives me an error.

xcube/util/jsonencoder.py
konstntokas: done

test/webapi/ows/stac/test_controllers.py
konstntokas: > Ok. However, you should keep track of the path and output it as part of the error message.

done

konstntokas: > But actually it is strange why this is needed. `TestCase.assertEquals()` actually does a deep equality comparison.

`TestCase.assertEqual()` also checks for the order of the keys in the dictionary, which gives me an error.

test/webapi/ows/stac/test_controllers.py
pont-us: I was wondering why you define `assert_equal_deep_dict` rather than just using `self.assertEqual()`, and eventually realized that you presumably need something that's insensitive to the order of iterables within the dictionaries. I suggest adding a short docstring to `assert_equal_deep_dict` to explain this.

pont-us (approved):
I had one suggestion (see comment), but approving since I don't need to review again.

test/webapi/ows/stac/test_controllers.py
konstntokas: > I was wondering why you define `assert_equal_deep_dict` rather than just using `self.assertEqual()`, and eventually realized that you presumably need something that's insensitive to the order of iterables within the dictionaries. I suggest adding a short docstring to `assert_equal_deep_dict` to explain this.

done

PR #1083, opened 2024-11-05
Allow to open netcdf file from remote HTTPS server

konstntokas:
This PR adjusts the `DatasetNetcdfFsDataAccessor` class, so that NetCDF files can be now opened from a remote HTTPS server. It adjusts the file uri to `http(s)://*#mode=bytes`, which allows to simply use `xarray.open_dataset` to open the data remotely. This also supports lazy when setting `chunks={}` or to a specific tile size, as used before. 

Checklist:

* [x] GitHub CI passes
* [ ] AppVeyor CI passes


test/core/store/fs/impl/test_https_netcdf.py
```diff
@@ -0,0 +1,43 @@
...
+    This class tests a lazy access of a NetCDF file from a remote HTTPS server.
+    """
+
+    def test_open_netcdf_https(self):
+        """This test loads GAMIv2-0_2010-2020_100m.nc (217GB), which is available via
+        https://datapub.gfz-potsdam.de/download/10.5880.GFZ.1.4.2023.006-VEnuo/
+        """
+        fs_path = "download/10.5880.GFZ.1.4.2023.006-VEnuo/GAMIv2-0_2010-2020_100m.nc"
+        store = new_data_store("https", root="datapub.gfz-potsdam.de")
+        ds = store.open_data(fs_path, chunks={})
```
konstntokas, line 29: This loads the data from remote source which is not wanted in a unittest. I would suggest using [pytest-recording](https://pypi.org/project/pytest-recording/), like in [xcube-stac](https://github.com/xcube-dev/xcube-stac?tab=readme-ov-file#some-notes-on-the-strategy-of-unit-testing-). This would mean that we add a new dependency pytest-recording to the test suite. Would that be okay? Or do you have a better suggestion?

konstntokas: @forman `pytest-recording` does not recognize the request when opening the data with `xr.open_dataset`. Therefore no cassette is written and it just uses the actual https request. I don't think we can use `pytest-recording` in this case. Probably we still need to build a mock, right?

konstntokas: I mocked the result of `xarray.open_dataset` now.

CHANGES.md
```diff
@@ -1,5 +1,9 @@
 ## Changes in 1.7.2 (in development)
 
+### Enhancements
+
+* The `DatasetNetcdfFsDataAccessor` class has been adjusted, so that NetCDF files can
+  be now opened from a remote HTTPS server using the `"https"` data store.
```
forman, line 6: Please take the perspective of a user. The user facing API in this case is the data store framework. How is it improved?

forman, line 6:
```suggestion
* The `"https"` data store (`store = new_data_store("https", ...)`) now allows 
  for lazily accessing NetCDF files.
  Implementation note: For this to work, the `DatasetNetcdfFsDataAccessor` 
  class has been adjusted.
```

forman (commented):
Good - but just see my suggestion.

You can merge otherwise.

test/core/store/fs/impl/test_https_netcdf.py
forman: Which is fine!

forman (approved):

PR #1095, opened 2024-12-16
Updated dataset spec

forman:
Changes in the doc:

- changed default name for CRS variable from `crs` to `spatial_ref`
- added spec how to define color encoding


AliceBalfanz (approved):
Thanks for the new details!

PR #1097, opened 2024-12-18
New data store preload API

### LINKED ISSUE #1093: Allow data stores to prepare accessing inert resources

Some data APIs provide only limited data access performance. Hence, the xcube data store's `store.open_data(data_id, ...)` approach doesn't apply well because (A) the method is blocking meaning it is not async, and (2) it may take long time before it returns. The reasons for data APIs to be slow are manifold: intended access is by file download, access is bandwidth limited, or not allowing for subsetting.

Give users the opportunity to prepare selected datasets for faster access. Therefore, the data store interface should be extended by a new method `preload_data()` that allows for preparing selected datasets for faster access. Of course, how this is done depends heavily on the specific data API. However, the data store framework could help implementing such preparations, e.g., by providing the means to download and cache data in generic filesystems (by the use of `fsspec`).
The input to the methods should be the data identifiers to be prepared, and data store specific preload parameters.
The return value of `preload_data()` should be a handle that represents the preloading process.
A second method should allow for describing allowed parameters passed to the `preload_data()`: `get_preload_params_schema()`. New method signarures:

```python
def preload_data(self, *data_ids: str, **preload_params: Any) -> PreloadHandle:
    """Preaload data for faster access."""
    return NullPreloadHandle()  # No effect

def get_preload_data_params(self) -> JsonObjectSchema:
    """Get the JSON schema that describes the parameters passed to `preload_data()`."""
    return JsonObjectSchema(additional_properties=False)
```

The `PreloadHandle` interface:

```python
@property
def is_cancelled() -> bool:
    """Whether the preload process has been cancelled."""

def cancel():
    """Cancel the preload process."""

@property
def is_closed() -> bool:
    """Whether the preload process has been closed."""

def close():
   """Close the preload process and release any bound resources.""""

@property
def progress() -> float:
   """Get the progress as a number in the range 0 to 1.""""

def subscribe(listener: Callable[[str, float, str], None]) -> Callable[[], None]:
   """Subscribe to state changes of the preload process.

   Args:
       listener: a listener that receives status, progress, and a message
   Returns: 
       A function to unsubscribe from the preload process.
    """
``` 

The `PreloadHandle` interface should also have an intuitive Jupyter Notebook representation to let user observe a long-running preload process.

forman:
- Enhanced the `xcube.core.store.DataStore` class to support _preloading_  of data resources. The preloading API is represented by new `xcube.core.store.DataPreloader` interface.
- Added handy default implementations `NullPreloadHandle` and `ExecutorPreloadHandle` to be returned by implementations of the `prepare_data()` method of a given data store.

Closes #1093

Checklist:

* [ ] GitHub CI passes
* [ ] AppVeyor CI passes

### DIFF
--- a/examples/notebooks/datastores/preload.ipynb
+++ b/examples/notebooks/datastores/preload.ipynb
@@ -1,0 +1,368 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "7f55e16a-46f3-4865-879a-a7cae151daa6",
+   "metadata": {},
+   "source": [
+    "### `ExecutorPreloadHandle` Demo\n",
+    "\n",
+    "This notebook is dedicated to developers wanting to enhance their\n",
+    "data store implementation by the new _data store preload API_. \n",
+    "This API has been added to `xcube.core.store.DataStore` in xcube 1.8.\n",
+    "Demonstrated here is the usage of the utility class ` xcube.core.store.preload.ExecutorPreloadHandle` \n",
+    "for cases where the preload process can be concurrently performed for each indiviual data resource. "
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "id": "7e27d98a-6519-4c4f-9e59-ae6369e102b9",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import random\n",
+    "import time\n",
+    "\n",
+    "from xcube.core.store.preload import ExecutorPreloadHandle\n",
+    "from xcube.core.store.preload import PreloadHandle\n",
+    "from xcube.core.store.preload import PreloadState\n",
+    "from xcube.core.store.preload import PreloadStatus"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "id": "92b57a80-14aa-4820-8acc-ab78fa45be0a",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "data_ids = (\n",
+    "    \"tt-data/tinky-winky.nc\", \n",
+    "    \"tt-data/dipsy.zarr\", \n",
+    "    \"tt-data/laa-laa.tiff\", \n",
+    "    \"tt-data/po.zarr.zip\"\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "id": "c3175fc6-084d-4da5-b7af-fe3f03758a7d",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def preload_data(handle: PreloadHandle, data_id: str):\n",
+    "    duration = random.randint(5, 15) # seconds\n",
+    "    num_ticks = 100  \n",
+    "    for i in range(num_ticks):\n",
+    "        time.sleep(duration / num_ticks)\n",
+    "        if handle.cancelled:\n",
+    "            # TODO: Note clear, why future.cancel() doesn't do the job\n",
+    "            handle.notify(PreloadState(data_id, status=PreloadStatus.cancelled))\n",
+    "            return\n",
+    "        handle.notify(PreloadState(data_id, progress=i / num_ticks))\n",
+    "        if i % 10 == 0:\n",
+    "            handle.notify(PreloadState(data_id, message=f\"Step #{i // 10 + 1}\"))\n",
+    "    handle.notify(PreloadState(data_id, progress=1.0, message=\"Done.\"))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "48792654-20e5-46db-aa47-37c03f46b0a7",
+   "metadata": {},
+   "source": [
+    "---\n",
+    "Synchronous / blocking call"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "id": "c8bb6812-af18-4e5f-8df0-d77889552e99",
+   "metadata": {},
+   "outputs": [
+    {
+     "data": {
+      "application/vnd.jupyter.widget-view+json": {
+       "model_id": "f88a78a4ffff41cd917f73405a1183e6",
+       "version_major": 2,
+       "version_minor": 0
+      },
+      "text/plain": [
+       "VBox(children=(HTML(value='<table>\\n<thead>\\n<tr><th>Data ID               </th><th>Status  </th><th>Progress …"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "handle = ExecutorPreloadHandle(data_ids, preload_data=preload_data)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "id": "8911ae1b-765c-4030-aeca-fc997e7a880d",
+   "metadata": {},
+   "outputs": [
+    {
+     "data": {
+      "text/html": [
+       "<table>\n",
+       "<thead>\n",
+       "<tr><th>Data ID               </th><th>Status  </th><th>Progress  </th><th>Message  </th><th>Exception  </th></tr>\n",
+       "</thead>\n",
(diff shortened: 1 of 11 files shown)


konstntokas (commented):
I like very much, that we already have the multi-threading in there.I think this makes it fairly easy to do the implementation in the data store plugins. 

I tested the notebook locally which works fine!

xcube/core/store/accessor.py
```diff
@@ -307,6 +323,62 @@ def write_data(
...
+    def preload_data(
+        self,
+        *data_ids: str,
+        **preload_params: Any,
+    ) -> PreloadHandle:
+        """Preload the given data items for faster access.
+
+        The method is blocking by default. You can pass a `monitor` to observe the
+        preload process, to cancel it, or to pass a function that is called if the
+        preload process successfully completed.
```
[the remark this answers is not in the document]
forman: Oops, this should no longer be in there. I removed `monitor` for time being!

xcube/core/store/preload.py
```diff
@@ -0,0 +1,404 @@
...
+        self.notify(PreloadState(data_id, status=PreloadStatus.started))
+        self.preload_data(data_id)
+        return data_id
+
+    def _handle_preload_data_done(self, f: Future[str]):
+        data_id: str | None = None
+        for data_id, future in self._futures.items():
+            if f is future:
+                break
+        if data_id is None:
```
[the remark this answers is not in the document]
forman: I search for the `data_id` that belongs to the passed future `f`.

I'll add a comment.

xcube/core/store/preload.py
```diff
@@ -0,0 +1,404 @@
...
+        return False
+
+    def cancel(self):
+        pass
+
+    def show(self) -> Any:
+        return None
+
+
+class ExecutorPreloadHandle(PreloadHandle):
```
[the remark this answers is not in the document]
forman: > The preload function then contains only the handling of download and processing of the data, while the `ExecutorPreloadHandle` class takes care of threading, visualization etc. Is that right?

Yupp.

xcube/core/store/accessor.py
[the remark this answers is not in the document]
forman: Fixed.

xcube/core/store/preload.py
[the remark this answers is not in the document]
forman: Done.

konstntokas (approved):
I just saw that the failed test are not related to preload API. So I approve :)

xcube/core/store/preload.py
```diff
@@ -0,0 +1,409 @@
...
+        Concurrently executes the *preload_data* passed to the constructor,
+        if any. Otherwise, it does nothing.
+
+        Can be overridden by clients to implement the actual preload operation.
+
+        Args:
+           data_id: The data identifier of the data resource to be preloaded.
+        """
+        if self._preload_data is not None:
+            self._preload_data(self, data_id)
```
b-yogesh, line 232: I am trying to understand why we have this method here. Seems redundant to me. In the method below `_run_preload_data`, we can directly call the preload method passed in.
So I have the following doubt:
- The role of this class is to handle the preload jobs such as updating the status, progress, message etc., cancelling the preload event and showing the user the overall status. Then why do we need this method here as it seems we are mixing responsibilities? The preload_data() function (not this method) that we pass in as a dependency should be responsible for how the data is preloaded and this class should be oblivious to it. I would be interested in knowing about the intuition behind having this method here.

xcube/core/store/preload.py
```diff
@@ -0,0 +1,409 @@
...
+
+    def __str__(self):
+        return ", ".join(f"{k}={v}" for k, v in _to_dict(self).items())
+
+    def __repr__(self):
+        args = ", ".join(f"{k}={v!r}" for k, v in _to_dict(self).items())
+        return f"{self.__class__.__name__}({args})"
+
+
+class PreloadHandle(ABC):
```
b-yogesh, line 77: We also talked about the `cache_data_store` and the `subscribe` method. Will this be added in the next PRs?

xcube/core/store/preload.py
```diff
@@ -0,0 +1,409 @@
...
+    def __enter__(self) -> "PreloadHandle":
+        """Enter the context.
+
+        Does nothing but returning this handle.
+        Only useful when in blocking mode.
+
+        Returns:
+            This object.
+        """
+        return self
```
b-yogesh, line 282: This is the same as the base class, can be removed.

examples/notebooks/datastores/preload.ipynb
```diff
@@ -0,0 +1,368 @@
+{
```
b-yogesh, line 1: We can also add the context manager example in the notebook.

b-yogesh (changes requested):
It looks good overall, thanks. I have added some comments below, please have a look.

xcube/core/store/preload.py
```diff
@@ -0,0 +1,409 @@
...
+        return None
+
+
+class ExecutorPreloadHandle(PreloadHandle):
+    """TODO - Add docs"""
+
+    def __init__(
+        self,
+        data_ids: tuple[str, ...],
+        preload_data: Callable[[PreloadHandle, str], None] | None = None,
```
konstntokas, line 170: Maybe it is useful to allow `preload_params` in the `preload_data` function. 

E.g. I had the case, where I had multiple tiff files within a zip file which could be merged if the user requests it by a keyword boolean.

xcube/core/store/preload.py
```diff
@@ -0,0 +1,409 @@
...
+        executor: Executor | None = None,
+        blocking: bool = True,
+    ):
+        self._preload_data = preload_data
+        self._executor = executor or ThreadPoolExecutor()
```
b-yogesh, line 175: I also wanted to discuss the number of threads that can be started concurrently which would vary according to the different machines that the users would use. We should maybe try to figure out what is the maximum number of threads to start also based on the tasks in the preload_data() function. Should this be done in the preload_data where the user defines the executor or we take care of that in the xcube interface, that is up for discussion.

xcube/core/store/preload.py
forman: See the docstring:

> Can be overridden by clients to implement the actual preload operation.

xcube/core/store/preload.py
forman: > Should this be done in the preload_data where the user defines the executor

Yes, clients can add their own `executor`.

> or we take care of that in the xcube interface

No.

xcube/core/store/preload.py
forman: The idea is to support 2 use cases:

- derive from `ExecutorPreloadHandle` and override `preload_data()` method
- use `ExecutorPreloadHandle` as-is and pass your own `preload_data` function

xcube/core/store/preload.py
forman: Yepp.

xcube/core/store/preload.py
forman: I don't understand the use case. Why don't you just add a dedicated opener parameter then?
You could also use `functools.partial()`.

xcube/core/store/preload.py
forman: - `cache_data_store` is an implementation detail
- `subscribe` can be added if needed. The current design and programming model

xcube/core/store/preload.py
```diff
@@ -0,0 +1,445 @@
...
+        self.message = message
+        self.exception = exception
+
+    def update(self, event: "PreloadState"):
+        """Update this state with the given partial state.
+
+        Args:
+            event: the partial state.
+        """
+        assert_instance(event, PreloadState, name="data_id")
```
b-yogesh, line 58: I think the `name` in the `assert_instance` should be `event` so that the assertion reads `event must be an instance of PreloadState, was something_else`
```suggestion
        assert_instance(event, PreloadState, name="event")
```

b-yogesh (approved):
Approved! Added a tiny change suggestion. Please have a look before merging!

PR #1098, opened 2024-12-20
Improve resample_in_space and rectify_dataset

konstntokas:
This PR improves the `resampling_in_space` method, if another CRS is required and therefore a reprojection needs to be done.  

* The method `xcube.core.GridMapping.transform` now supports lazy execution. If
  computations based on actual data are required—such as determining whether the
  grid mapping is regular or estimating the resolution in the x or y direction—only a
  single chunk is accessed whenever possible, ensuring faster performance.
-> see large example `examples/notebooks/resampling/reproject_large_esa_cci_landcover.ipynb` execution time of cell 5 drops from 1.42min to ~800ms.

* The function `xcube.core.resampling.rectify_dataset` now supports `xarray.Datasets`
  containing multi-dimensional data variables structured as `var(..., y_dim, x_dim)`.
  The two spatial dimensions (`y_dim` and `x_dim`) must occupy the last two positions 
  in the variable's dimensions.
-> see large example `examples/notebooks/resampling/reproject_large_esa_cci_landcover.ipynb` execution time of cell 6 drops from 4.51min to ~2.16ms, even though now 5 time steps are selected compared to just one. 

Checklist:

* [x] Add unit tests and/or doctests in docstrings
* [ ] ~Add docstrings and API docs for any new/modified user-facing classes and functions~
* [ ] ~New/modified features documented in `docs/source/*`~
* [x] Changes documented in `CHANGES.md`
* [x] GitHub CI passes
* [x] AppVeyor CI passes
* [x] Test coverage remains or increases (target 100%)

PR #1100, opened 2025-01-03
Minor adjustments and bug fixing of resampling_in_space

konstntokas:
This PR is related to the previous PR #1098. Further I enhanced the testing of `resample_in_space` to cover the case of a reprojection, where a subsetting in `rectify_dataset` is performed.  

Checklist:

* [x] GitHub CI passes
* [x] AppVeyor CI passes

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -55,6 +55,9 @@
 
 ### Fixes
 
+* The function `xcube.core.resample.resample_in_space()` now supports the parameter
+  `source_ds_subset=True` when calling `rectify_dataset`. This feature enables
+  performing the reprojection exclusively on the congruent subset of the dataset.
 * The function `xcube.core.resample.resample_in_space()` now always operates
   lazily and therefore supports chunk-wise, parallel processing. (#1082)
 * Bux fix in the `has_data` method of the `"https"` data store
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -57,7 +57,8 @@
 
 * The function `xcube.core.resample.resample_in_space()` now supports the parameter
   `source_ds_subset=True` when calling `rectify_dataset`. This feature enables
-  performing the reprojection exclusively on the congruent subset of the dataset.
+  performing the reprojection exclusively on the spatially congruent subset of 
+  the dataset.
 * The function `xcube.core.resample.resample_in_space()` now always operates
   lazily and therefore supports chunk-wise, parallel processing. (#1082)
 * Bux fix in the `has_data` method of the `"https"` data store
--- a/xcube/core/gridmapping/cfconv.py
+++ b/xcube/core/gridmapping/cfconv.py
@@ -72,7 +72,6 @@ def get_dataset_grid_mapping_proxies(
     grid_mapping_proxies: dict[Union[Hashable, None], GridMappingProxy] = dict()
 
     # Find any grid mapping variables by CF 'grid_mapping' attribute
-    #
     for var_name, var in dataset.variables.items():
         grid_mapping_var_name = var.attrs.get("grid_mapping")
         if (
@@ -86,7 +85,6 @@ def get_dataset_grid_mapping_proxies(
 
     # If no grid mapping variables found,
     # try if CRS is encoded in some variable's attributes
-    #
     if not grid_mapping_proxies:
         for var_name, var in dataset.variables.items():
             gmp = _parse_crs_from_attrs(var.attrs)
@@ -96,15 +94,12 @@ def get_dataset_grid_mapping_proxies(
 
     # If no grid mapping variables found,
     # try if CRS is encoded in dataset attributes
-    #
     if not grid_mapping_proxies:
         gmp = _parse_crs_from_attrs(dataset.attrs)
         if gmp is not None:
             grid_mapping_proxies[None] = gmp
 
     # Find coordinate variables.
-    #
-
     latitude_longitude_coords = GridCoords()
     rotated_latitude_longitude_coords = GridCoords()
     projected_coords = GridCoords()
@@ -112,7 +107,6 @@ def get_dataset_grid_mapping_proxies(
     potential_coord_vars = _find_potential_coord_vars(dataset)
 
     # Find coordinate variables that use a CF standard_name.
-    #
     coords_standard_names = (
         (latitude_longitude_coords, "longitude", "latitude"),
         (rotated_latitude_longitude_coords, "grid_longitude", "grid_latitude"),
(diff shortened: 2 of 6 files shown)


xcube/core/resampling/rectify.py
```diff
@@ -138,7 +138,18 @@ def rectify_dataset(
...
+            # `resample_in_space` function.
+            if "transformed_x" and "transformed_y" in source_ds_subset:
+                source_gm = GridMapping.from_coords(
+                    source_ds_subset.transformed_x,
+                    source_ds_subset.transformed_y,
+                    crs=source_gm.crs,
+                    tile_size=source_gm.tile_size,
+                )
+            else:
+                source_gm = GridMapping.from_dataset(source_ds_subset)
```
TonioF, line 152: This works, but I have the feeling this is a workaround. As the coordinates are included in source_ds, it should be possible to call 

GridMapping.from_dataset(source_ds_subset, prefer_crs=source_gm.crs) 

and get the correct result.

konstntokas: I changed the setup up the `source_ds` in `resample_in_space`. I took Sentinel-3 data as inspiration (see notebook https://github.com/xcube-dev/xcube/blob/main/examples/notebooks/resampling/rectify_dataset.ipynb).

TonioF (changes requested):
The new solution would cause problems with data variables that have a gridmapping attribute that points to a crs variable which might not exist anymore after these changes. I still prefer the solution of a variable "transformed_spatial_ref". However, if you prefer your way, please make sure the attributes are not wrong.

konstntokas:
> The new solution would cause problems with data variables that have a gridmapping attribute that points to a crs variable which might not exist anymore after these changes. I still prefer the solution of a variable "transformed_spatial_ref". However, if you prefer your way, please make sure the attributes are not wrong.

Thank you that you are looking into it so carefully!

I need to remove the original coordinates, because if the source ds has `x` and `y`, then either `x` and `y` or `transformed_x` and `transformed_y` will be added to the `projected_coords` in the function `xcube.core.gridmapping.cfconv.get_dataset_grid_mapping_proxis`, depending if `x` or `transformed_x` (same for y) comes first in the list `potential_coord_vars`. I therefore prefer to setup a clean dataset in `resample_in_space`, where only one set of coordinates can be found. 

I thus would like to keep my suggestion and also delete all grid_mappings from the data variables in the source_ds. I also added a `grid_mapping` attribute to the unittest.

konstntokas:
> > I need to remove the original coordinates, because if the source ds has `x` and `y`, then either `x` and `y` or `transformed_x` and `transformed_y` will be added to the `projected_coords` in the function `xcube.core.gridmapping.cfconv.get_dataset_grid_mapping_proxis`, depending if `x` or `transformed_x` (same for y) comes first in the list `potential_coord_vars`. I therefore prefer to setup a clean dataset in `resample_in_space`, where only one set of coordinates can be found.
> 
> Do you really have to? For testing, I kept existing variables crs and spatial_ref and added a new one transformed_spatial_ref. Now, even though the source_gm retrieved from GridMapping.from_dataset() was the UTM one, the tests eventually succeeded.

I changed it to `transformed_spatial_ref` without deleting the previous crs or spatail_ref variable/coordinate. The the CI fails in the newly added test, if actual subsetting is performed, see second test in `test_spatial.test_reproject_utm()`.

TonioF:
You're right, when I executed this, I had left in the line where x and y were removed from the dataset. I now would accept your solution from before the last two commits. It's still not perfect, but I wouldn't be able to come up with a better solution.

konstntokas:
@forman can you have a look onto our discussion and my suggested changes. I think a second opinion is useful here. Thank you!

CHANGES.md
```diff
@@ -55,6 +55,9 @@
 
 ### Fixes
 
+* The function `xcube.core.resample.resample_in_space()` now supports the parameter
+  `source_ds_subset=True` when calling `rectify_dataset`. This feature enables
+  performing the reprojection exclusively on the congruent subset of the dataset.
```
forman, line 60:
```suggestion
  performing the reprojection exclusively on the spatially congruent subset of 
  the dataset.
```

xcube/core/resampling/spatial.py
```diff
@@ -208,11 +208,33 @@ def resample_in_space(
     # If CRSes are not both geographic and their CRSes are different
     # transform the source_gm so its CRS matches the target CRS:
     transformed_source_gm = source_gm.transform(target_gm.crs, xy_res=target_gm.xy_res)
+    source_ds = source_ds.drop_vars(source_gm.xy_dim_names)
+    if "crs" in source_ds:
+        source_ds = source_ds.drop_vars("crs")
+    if "spatial_ref" in source_ds:
+        source_ds = source_ds.drop_vars("spatial_ref")
```
forman, line 215: This may not be right in cases. You should just drop the grid mapping variable, no matter how it is named. `crs` and `spatial_ref` are common names, but their names are not a sufficient condition to identify the grid mapping variable. There can even be more than one such variables if multiple CRSes are used.

xcube/core/gridmapping/cfconv.py
```diff
@@ -72,7 +72,6 @@ def get_dataset_grid_mapping_proxies(
     grid_mapping_proxies: dict[Union[Hashable, None], GridMappingProxy] = dict()
 
     # Find any grid mapping variables by CF 'grid_mapping' attribute
-    #
```
forman, line 75: I like these spacers so the comment is not glued to the next line, but still connected.

CHANGES.md
forman: Not sure, maybe _congruent_ applies only to the spatial domain. Then ignore my suggestion.

xcube/core/resampling/spatial.py
konstntokas: I now go through the `grid_mapping` in the data variable attributes, and  not only delete the attribute with key key `grid_mapping`, but also drop the grid mapping variable. Afterwards I still check for the grid mapping variable named with `crs` and `spatial_ref`. For Sentinel 2 jp2 files e.g. the dataset has the coordinate `spatial_ref`, but the data variable has no attribute `grid_mapping`.

xcube/core/gridmapping/cfconv.py
konstntokas: No preference here. I changed it back. :)

xcube/core/resampling/spatial.py
forman: Thats good.

forman (approved):
@TonioF @konstntokas 

I approve unless I overlooked the place where you added `transformed_spatial_ref`. I'd be against it. Instead replace the existing source grid mapping from. If you need a history, place a record into the global `history` attribute.

PR #1133, opened 2025-03-06
PreloadDataStore added

### LINKED ISSUE #1132: The preload method shall return the cache data store
In xcube-zenodo, one preload data ID can contain multiple files, which corresponds to multiple data IDs in the cache data store. Thus there might  want a unique mapping between data ids in preload and open the data. 

To not confuse the user, the preload method shall return the cache store, which allows full functionality of accessing all preloaded data. The preload handle shall be included in the cache store.

konstntokas:
Closes #1132 

Checklist:

* [ ] GitHub CI passes
* [ ] AppVeyor CI passes

### DIFF
--- a/xcube/core/store/store.py
+++ b/xcube/core/store/store.py
@@ -487,7 +487,7 @@ class DataStore(DataOpener, DataSearcher, DataPreloader, ABC):
         self,
         *data_ids: str,
         **preload_params: Any,
-    ) -> PreloadHandle:
+    ) -> "PreloadDataStore":
         """Preload the given data items for faster access.
 
         Warning: This is an experimental and potentially unstable API
@@ -512,9 +512,10 @@ class DataStore(DataOpener, DataSearcher, DataPreloader, ABC):
             A handle for the preload process. The default implementation
             returns an empty preload handle.
         """
-        return NullPreloadHandle()
-
-    # noinspection PyMethodMayBeStatic
+        cache_store = new_data_store("file")
+        cache_store.preload_handle = NullPreloadHandle()
+        assert isinstance(cache_store, PreloadDataStore)
+        return cache_store
 
 
 class MutableDataStore(DataStore, DataWriter, ABC):
@@ -690,3 +691,22 @@ class MutableDataStore(DataStore, DataWriter, ABC):
         Raises:
             DataStoreError: If an error occurs.
         """
+
+
+class PreloadDataStore(MutableDataStore):
+    """A preload data store is a multable data store which contains the preload handle.
+
+    Instances of this class are returned by the ``DataStore.preload_data()`` method.
+    """
+
+    @property
+    @abstractmethod
+    def preload_handle(self) -> PreloadHandle:
+        """This field must be defined and set in subclasses."""
+        pass
+
+    @preload_handle.setter
+    @abstractmethod
+    def preload_handle(self, value: PreloadHandle) -> None:
+        """Setter for preload_handle."""
+        pass
--- a/xcube/core/store/preload.py
+++ b/xcube/core/store/preload.py
@@ -75,11 +75,7 @@ class PreloadState:
 
 
 class PreloadHandle(ABC):
-    """A handle for a preload job.
-
-    Instances of this class are returned by the
-    ``DataStore.preload_data()`` method.
-    """
+    """A handle for a preload job."""
 
     @abstractmethod
     def get_state(self, data_id: str) -> PreloadState:
--- a/xcube/util/cache.py
+++ b/xcube/util/cache.py
@@ -439,9 +439,7 @@ def _compute_object_size(obj):
                 else (
                     3
                     if m in ("RGB", "YCbCr", "LAB", "HSV")
-                    else 1.0 / 8.0
-                    if m == "1"
-                    else 1
+                    else 1.0 / 8.0 if m == "1" else 1
                 )
             )
         )
--- a/xcube/util/dask.py
+++ b/xcube/util/dask.py
@@ -218,9 +218,7 @@ def new_cluster(
     cluster_account = (
         account
         if account is not None
-        else account_from_env_var
-        if account_from_env_var is not None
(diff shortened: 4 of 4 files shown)


xcube/core/store/store.py
```diff
@@ -509,12 +509,13 @@ def preload_data(
...
-            returns an empty preload handle.
+            A mutable data store containing the preload handle.
+            The default implementation contains an empty preload handle.
         """
-        return NullPreloadHandle()
-
-    # noinspection PyMethodMayBeStatic
+        cache_store = new_data_store("file")
+        cache_store.preload_handle = NullPreloadHandle()
+        assert isinstance(cache_store, PreloadDataStore)
```
konstntokas, line 517: I do not like this, but I do not know how to force `cache_store` to be a  `PreloadDataStore`

forman: I don't like this either :)

The default you return here should be a dataset that wraps this (`self`) data store and that returns a `NullPreloadHandle()` as its preload handle.

xcube/core/store/store.py
```diff
@@ -690,3 +691,22 @@ def deregister_data(self, data_id: str):
         Raises:
             DataStoreError: If an error occurs.
         """
+
+
+class PreloadDataStore(MutableDataStore):
```
forman, line 696: Does it really need to be a `MutableDataStore`? Note that *mutable* refers to the user's perspective. Should users mutate the returned preload store? I suppose not.

xcube/core/store/store.py
```diff
@@ -690,3 +691,22 @@ def deregister_data(self, data_id: str):
...
+    """A preload data store is a multable data store which contains the preload handle.
+
+    Instances of this class are returned by the ``DataStore.preload_data()`` method.
+    """
+
+    @property
+    @abstractmethod
+    def preload_handle(self) -> PreloadHandle:
+        """This field must be defined and set in subclasses."""
```
forman, line 705: Please explain the property's purpose rather than saying that the property is abstract - that's already indicated by the `@abstractmethod` decorator. You can tell developers what is expected here.

```suggestion
        """The preload handle that can be used to observe the preload progress.
        Implementors of this interface may use a `DefaultProgressHandle` or consider
        returning a `NullProgressHandle` if the progress is not observable.        
        """
```

xcube/core/store/store.py
```diff
@@ -690,3 +691,22 @@ def deregister_data(self, data_id: str):
...
+        pass
+
+    @preload_handle.setter
+    @abstractmethod
+    def preload_handle(self, value: PreloadHandle) -> None:
+        """Setter for preload_handle."""
+        pass
```
forman, line 712: This should not be part of the interface. Users should not change the handle. The handle is set by data store implementors, not users.

forman (changes requested):

xcube/core/store/store.py
konstntokas: Ah I though the returned store shall be used for writing. But this makes total sense, what you say. The preload data store should only be there to access the data being preloaded.

xcube/core/store/store.py
konstntokas: changes slightly as the classes are called NullPreloadHandle and ExecutorPreloadHandle in the preload module.

xcube/core/store/store.py
konstntokas: But if I remove `assert isinstance(cache_store, PreloadDataStore)`, I get a warning in the code checking: Expected type 'PreloadDataStore', got 'DataStore' instead. The class type is not adapted dynamically.

xcube/core/store/store.py
forman: No. `cache_store` should not be exposed at all, it is an implementation detail of your Zenodo store. Of course it can be mutable, but the _interface_ that users receive from calling `preload_data()` should not expose muability. Hence the interface `PreloadDataStore` should not be mutable.

Btw, I also just realize that it should be named `PreloadedDataStore` as it is a data store for _preloaded_ data.

xcube/core/store/__init__.py
```diff
@@ -38,6 +38,7 @@
 from .store import (
     DataStore,
     MutableDataStore,
+    PreloadDataStore,
```
forman, line 41: Rename to `PreloadedDataStore`, see comment below.

forman (changes requested):

xcube/core/store/__init__.py
konstntokas: Done

xcube/core/store/store.py
```diff
@@ -690,3 +689,19 @@ def deregister_data(self, data_id: str):
...
+    """
+
+    @property
+    @abstractmethod
+    def preload_handle(self) -> PreloadHandle:
+        """The preload handle that can be used to observe the preload progress.
+        Implementors of this interface may use a `ExecutorPreloadHandle` or consider
+        returning a `NullPreloadHandle` if the progress is not observable.
+        """
+        pass
```
forman, line 707: Remove line. `pass` is redundant because we have a docstring which acts as no-op statement.

xcube/core/store/store.py
```diff
@@ -509,12 +509,11 @@ def preload_data(
...
-            A handle for the preload process. The default implementation
-            returns an empty preload handle.
+            A mutable data store containing the preload handle.
+            The default implementation contains an empty preload handle.
         """
-        return NullPreloadHandle()
-
-    # noinspection PyMethodMayBeStatic
+        self.preload_handle = NullPreloadHandle()
+        return self
```
forman, line 516: Add unit test that asserts we receive `self` and that asserts a `NullPreloadHandle` is used.

forman (changes requested):

xcube/core/store/store.py
konstntokas: I added a unittest for this. :)

PR #1136, opened 2025-03-11
Add Demo Panel - Spectral View

clarasb:
this PR uses: https://github.com/xcube-dev/xcube-viewer/pull/495

This PR creates a 'Spectral View'-plot:

- for one selected Place (Point)
- at a selected time 
- only with Variables that represent wavelengths

in the future:
- for selected Places
- for selected Variables

grafik

Checklist:

* [ ] GitHub CI passes
* [ ] AppVeyor CI passes


examples/serve/panels-demo/my_viewer_ext/my_panel_3.py
```diff
@@ -0,0 +1,320 @@
...
+
+# TODO implement this more efficiently
+@panel.callback(
+    Output("time", "children"),
+)
+def update_last_timelabel(
+    new_timelabel: str,
+) -> list[str]:
+
+    return [new_timelabel]
```
b-yogesh, line 320: Hard to understand what you are doing here as you not defined `time` correctly.

b-yogesh (changes requested):
There are several problems with this PR
1. Panel is not rendered right (the dropdowns and update button is rendered at the bottom of the panel)
2. Chart can be created even without selecting any options from the dropdown, and this should not be happening.
3. Chart is not updated when different variables are selected. Either that variable is not used for rendering the chart or the logic does not handle the new variable. In either case, it should be fixed.
4. `place_group` has different types all over the different functions -> it should
 stay the same `list[dict[str,Any]]`
5. If there are multiple geometries, this code does not filter for Point which is
 what the PR says
6. After generating the chart, when I click on settings, the chart disappears.
7. Default view of this panel shows error (list index out of range) when no point is selected initially.
8. You use `Component` which is a base class and should not be used. Only the derived components from this class are available to be used.
9. This is an assumption, but if we have multiple points, I assumed you implemented to add all of them on the same chart, because currently only supports one point.
10. Add type annotations for every parameter in every function defined consistently.

examples/serve/panels-demo/my_viewer_ext/my_panel_3.py
clarasb: after some changes in the code, this function became obsolete and got deleted.

clarasb:
I worked on the points from above

1. solved
2. this is depended on the newest charlets release
3. depended on the newest charlets release.
4. solved
5. solved
6. solved
7. solved, there is no error shown anymore
8. got it, this feature is solved differently now
9. yes, this features is also depended on the multiple selection (new chartlets release)
10. solved

Some features of the Panel are only available with the newest chartlets release. Until the release, the features got excluded from the panel. See screenshot for current available features of Panel.
These are the features:
- selection of multiple points
- selection of Variables of interest

grafik

b-yogesh (commented):
LGTM. Just one tiny change requested.

b-yogesh (approved):

PR #1142, opened 2025-03-31
Allow using other xarray backends

forman:
In this PR:

- Allow using other xarray backends than "zarr" for Zarr datasets.
- When computing dataset bounding boxes, respect that 64-bit integer type coordinates must explicitly be cast to float to avoid raising TypeErrors.

I couldn't find a good way to test the new code in `xcube/core/store/fs/impl/dataset.py` as it would require a new additional xarray backend registered temporary.

### DIFF
--- a/test/core/store/fs/test_registry.py
+++ b/test/core/store/fs/test_registry.py
@@ -5,6 +5,7 @@
 import collections.abc
 import os.path
 import shutil
+import tempfile
 import unittest
 import warnings
 from abc import ABC, abstractmethod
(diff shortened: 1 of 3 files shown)


konstntokas (changes requested):
Please see my comment regarding `get_dataset_bounds`

konstntokas (approved):
It is not wrong, but from my understanding it is enough to only change 

```
x_min -= 0.5 * delta
x_max += 0.5 * delta
```
to
```
x_min = x_min - 0.5 * delta
x_max = x_max + 0.5 * delta
```
All the `float` changes are not necessary. 

Anyhow it is not wrong, so I do not want to block you,

PR #1152, opened 2025-04-14
Add reprojection

konstntokas:
This PR adds `xcube.core.resampling.reproject_dataset`, a new method for reprojecting datasets to a different coordinate reference system. It offers improved computational efficiency and robustness compared to `xcube.core.resampling.resample_in_space`.

Checklist:

* [x] GitHub CI passes
* [x] AppVeyor CI passes


CHANGES.md
```diff
@@ -4,6 +4,13 @@
 
 * Make test suite compatible with click >=8.2.0 (#1155)
 
+### Enhancements
+
+* Added `xcube.core.resampling.reproject_dataset`, a new method for reprojecting 
+  datasets to a different coordinate reference system. It offers improved 
+  computational efficiency and robustness compared to 
+  `xcube.core.resampling.resample_in_space`.
```
forman, line 12: Questions that immediately pop up when reading this is: Why is `resample_in_space` still there? Why haven't they just improved `resample_in_space`? When should I use `resample_in_space` and when `reproject_dataset`? 

Please answer such questions in the change description and the docstring.

xcube/core/resampling/affine.py
```diff
@@ -158,12 +158,18 @@ def resample_dataset(
                 aggregator = np.nanmean
                 # forman: changed default from True to False (v1.6, 2024-06-05)
                 recover_nan = False
+            if hasattr(var.data, "chunksize"):
+                chunksize = list(var.data.chunksize)
+                chunksize[-2] = tile_height
+                chunksize[-1] = tile_width
+            else:
+                chunksize = (tile_height, tile_width)
```
forman, line 166: This is a surprising behaviour. Parameters explicitly passed into a function should be considered or the user should be informed by raising or warning. If `tile_size` is `None`, _then_ you can fallback to default values derived from internal data chunking.

xcube/core/resampling/reproject.py
```diff
@@ -0,0 +1,493 @@
...
+        Default is ``numpy.nanmean`` for floating point variables,
+        and None (= nearest neighbor) for integer and bool variables.
+    * ``recover_nan`` (bool) - whether a special algorithm
+        shall be used that is able to recover values that would
+        otherwise yield NaN during resampling.
+        Default is False for all variable types since this
+        may require considerable CPU resources on top.
+
+
+    Ars:
```
forman, line 69:
```suggestion
    Args:
```

xcube/core/resampling/reproject.py
```diff
@@ -0,0 +1,493 @@
...
+def reproject_dataset(
+    source_ds: xr.Dataset,
+    source_gm: GridMapping | None = None,
+    target_gm: GridMapping | None = None,
+    ref_ds: xr.Dataset | None = None,
+    fill_value: int | float | None = None,
+    interpolation: str | None = "nearest",
+    var_configs: Mapping[Hashable, Mapping[str, Any]] | None = None,
+):
+    """
```
forman, line 36: Please 
- Say, why this function exists, when there is also `resample_in_space`.
- Tell, when to use which one, e.g. pros/cons.
- Change docstring of `resample_in_space` to mention this function and when to use it instead.

xcube/core/resampling/reproject.py
```diff
@@ -0,0 +1,493 @@
...
+            - other integers: -1
+        interpolation: Optional. Interpolation method to use. Must be one of:
+            "nearest", "triangular", or "bilinear". Defaults to "nearest".
+            "Triangular" uses 3 adjacent pixels, "bilinear" uses 4. These methods
+            apply only to floating-point data. Convert integer data to float if
+            interpolation is needed.
+        downscale_var_configs: Optional resampling configurations
+            for individual variables applied during downscaling.
+
+    Returns: A reprojected dataset.
```
forman, line 89: Provide more details of the returned dataset, e.g., the grid mapping variable.

forman (changes requested):
This is part 1 of 2 of my review. Please address my points, then in part 2 I'll study the algorithm carefully.

CHANGES.md
konstntokas: done

xcube/core/resampling/affine.py
konstntokas: `tile_size` comes from `target_gm.tile_size` in line 96, and this is never `None`. Here, I have not changes a lot. I only cover the case for a 3D datacube, where aggregation is needed in the reprojection process. Then I use `affine_transform_dataset`, which without these changes would change my chunking along the 3rd axis to 1, even though the chunksize beforehand was larger. The else case is needed for `var.data` being a numpy array.

xcube/core/resampling/reproject.py
konstntokas: Done

xcube/core/resampling/reproject.py
konstntokas: done

xcube/core/resampling/reproject.py
konstntokas: done

xcube/core/resampling/affine.py
forman: Ok, that's fine. However, then we should pass a flag to indicate that the tile size was not user provided. I just want to avoid the case where `tile_size` supplied by the user is silently ignored.

forman: Ok!

xcube/core/resampling/reproject.py
```diff
@@ -0,0 +1,501 @@
...
+        else:
+            ds_out[var_name] = (
+                (data_array.dims[0], y_name, x_name),
+                array_reprojected,
+            )
+        ds_out[var_name].attrs = data_array.attrs
+    return ds_out
+
+
+def _reproject_block(
```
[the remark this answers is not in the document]
forman: Tried it, but unfortunately it does not yet support multi-dim indexes as used in 

```
        ix = np.rint(ix).astype(np.int16)
        iy = np.rint(iy).astype(np.int16)
        data_reprojected = scr_data[:, iy, ix]
```

forman (changes requested):
Consider it as expected if you address my minor comments and suggestions. Great work!

I have not fully understood the implementation yet, for this it would be great to have a more detailed description of the algorithm in the user docs = another PR.

xcube/core/resampling/reproject.py
```diff
@@ -0,0 +1,501 @@
...
+
+        The methods `reproject_dataset` and `resample_in_space` produce nearly identical
+        results when reprojecting to a different CRS, with only negligible differences.
+        `resample_in_space` remains available to preserve compatibility with existing
+        services. Once `reproject_dataset` proves stable in production use, it may be
+        integrated into `resample_in_space`.
+    """
+
+    # translate interpolation mode
+    interpolation_mode = _INTERPOLATIONS.get(interpolation or "nearest")
```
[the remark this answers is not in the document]
konstntokas: I added a verification of `interpolation_mode` being a `ValueError`. 

Regarding the fall back to `"bilinear"` in `_reproject_block()`, I just mimicked the same strategy as in `rectify_dataset`. Is true for for the function itself. But it is an internal function, and `interpolation_mode` is assigned by `interpolation_mode = _INTERPOLATIONS.get(interpolation or "nearest")`, so from outside, only "nearest", "triangular", "bilinear" or None is allowed, and if it is None, "nearest" will be assigned.

forman (approved):
Thanks for this great contribution!

Issue #726, opened 2022-09-14
Add function to list file extensions recognized by FS stores


pont-us:
**Is your feature request related to a problem? Please describe.**

I would like a quick and easy way to find out what file extensions are recognized by currently installed xcube stores.

**Describe the solution you'd like**

I would like a function that works something like this:

```python
>>> xcube.core.store.fs.get_known_file_extensions()

{
    '.zarr': 'zarr',
    '.nc': 'netcdf',
    '.tiff': 'geotiff'
    # etc.
}
```

**Describe alternatives you've considered**

I believe the most practical current alternative is to look at `xcube.core.store.fs.store._FILENAME_EXT_TO_FORMAT`, which contains this information, but the leading underscore indicates that this is only intended for internal use, and it's marked with a `TODO` indicating that it might be changed soon.

forman:
I'm thinking of introducing data identifiers extensions as a general data store concept. This would apply to stores that can automatically detect formats and their appropriate default in-memory models from identifier extensions, e.g. `.zarr`.

PR #1176, opened 2025-07-25
List filename extensions recognized by data stores

TonioF:
Addresses #726 . Also, added function get_filename_extensions to FsDataStore. This pr brought with it the dissolution of explicit mentions of formats and extensions in xcube.core.store.fs.store.py . These are now read from the entries for data openers and data writers in the extension registry. Also, this came with a revision of the way these are registered in plugin.py. Accessors now include in their metadata (a) what file extensions they support and (b) whether they are primary or secondary accesors. This information is useful to state that, e.g., a levels opener should open a levels dataset as a multileveldataset by default.

Checklist:

* [x] GitHub CI passes
* [x] AppVeyor CI passes


forman (approved):
Thanks :)

PR #1213, opened 2026-04-30
Extend WebAPI to derive time series and statistics for datasets with multiple non-spatial dimensions     

clarasb:
This PR extends `webapi` to support deriving statistics and time series for variables that have more dimensions than just `time`, `lat`, and `lon`.

- xcube Server now accepts and processes additional non-spatial dimensions (e.g. `depth`) when computing statistics and time series. Example request:

```
  GET /statistics/kattegat/chl?time=2025-07-15T12:00:00Z&depth=0.5016462206840515
```
- Removed special-cased handling of the `time` label. Now all `non_spatial_dimensions` are handled in the same way. If no explicit value is provided for a dimension, a value is selected automatically via `get_non_spatial_labels()`-
- `xcube.core.tile.get_non_spatial_labels` changed to a public function to support the default-value selection logic

Checklist:

* [x] GitHub CI passes
* [x] AppVeyor CI passes


xcube/webapi/timeseries/routes.py
```diff
@@ -77,6 +84,13 @@ async def post(self, datasetId: str, varName: str):
         end_date = self.request.get_query_arg(
             "endDate", type=pd.Timestamp, default=None
         )
+        non_spatial_dimensions = get_non_spatial_dimensions(
+            self.ctx, self.request, datasetId, varName
+        )
+        non_spatial_dimensions = {
+            k: v for k, v in non_spatial_dimensions.items() if k != "time"
+        }
```
forman: You could also enhance `get_non_spatial_dimensions()` by an opt. keyword argument `excluded_dims: Iterable[str] | None = None` to `. So the lines above would read:

```suggestion
        non_spatial_dimensions = get_non_spatial_dimensions(
            self.ctx, self.request, datasetId, varName, excluded_dims=["time"],
        )       
```

xcube/webapi/timeseries/routes.py
```diff
@@ -89,8 +103,25 @@ async def post(self, datasetId: str, varName: str):
...
+    variable = ds[var]
+
+    variable_dims = variable.dims
+    dimensions = {}
+    for dim in variable_dims:
+        value = request.get_query_arg(str(dim), type=str, default=None)
+        if value is not None:
+            dimensions[str(dim)] = value
+
+    labels = get_non_spatial_labels(ds, variable, labels=dimensions, logger=_logger, ignore=["time"])
```
forman: Oh, but it seems "time" is already excluded here!

forman (approved):
Nice! But see my comments!

PR #1216, opened 2026-06-16
Fix tile rendering

### LINKED ISSUE #605: Tiles are shifted by half a cell in x-and y-direction

clarasb:
## Summary 

Fix a half-pixel shift in raster tile rendering.

## Changes

- Sample tile coordinates at pixel centers instead of tile edges.
- Use tile width/height, not `width - 1` / `height - 1`, to compute tile pixel resolution.
- Round transformed dataset indices with `np.rint()` so rendered pixels use the nearest source cell.

## Reason

The previous coordinate generation placed tile samples on the outer tile bounds, which shifted rendered raster data by half a pixel. Sampling at pixel centers aligns the tile image with the intended raster footprint.

## Testing

- Updated tile test expectations for the corrected center-sampled output.

Closes https://github.com/xcube-dev/xcube-viewer/issues/605.

Checklist:

* [ ] ~~Add unit tests and/or doctests in docstrings~~
* [x] GitHub CI passes
* [x] AppVeyor CI passes


xcube/core/tile.py
```diff
@@ -194,18 +194,21 @@ def compute_tiles(
...
+        full_ds_dx = abs(float(full_ds_x_coords[1] - full_ds_x_coords[0]))
+        full_ds_dy = abs(float(full_ds_y_coords[1] - full_ds_y_coords[0]))
+
         num_extra_pixels = tile_enlargement
-        res_x = (ds_x_max - ds_x_min) / tile_width
-        res_y = (ds_y_max - ds_y_min) / tile_height
-        extra_dx = num_extra_pixels * res_x
-        extra_dy = num_extra_pixels * res_y
+        extra_dx = (num_extra_pixels + 0.5) * full_ds_dx
+        extra_dy = (num_extra_pixels + 0.5) * full_ds_dy
```
TonioF: I'm not convinced these changes are needed. I found that tests return the same result when the tile_enlargement parameter is left on its default value 1.

forman: I agree with @TonioF. Either justify the offset and add a comment or leave it alone.

xcube/core/tile.py
```diff
@@ -141,8 +141,8 @@ def compute_tiles(
         ds_y_coords = variable_0[ds_y_name]
         ds_y_points_up = bool(ds_y_coords[0] < ds_y_coords[-1])
 
-        tile_res_x = (tile_x_max - tile_x_min) / (tile_width - 1)
-        tile_res_y = (tile_y_max - tile_y_min) / (tile_height - 1)
+        tile_res_x = (tile_x_max - tile_x_min) / tile_width
+        tile_res_y = (tile_y_max - tile_y_min) / tile_height
```
forman, line 145: Omg, good catch! I really wonder why the `- 1` was introduced at all?!

forman (changes requested):
Great! Just address @TonioF and mine comment, then I'll approve.

clarasb:
@TonioF and @forman, thanks for your review. I reversed the mentioned changes and adjusted the affected tests.

xcube/core/tile.py
```diff
@@ -201,14 +201,20 @@ def compute_tiles(
...
+        extra_dy = num_extra_pixels * res_y
+
+        # full_ds_x_coords = variable_0[ds_x_name]
+        # full_ds_y_coords = variable_0[ds_y_name]
+        # full_ds_dx = abs(float(full_ds_x_coords[1] - full_ds_x_coords[0]))
+        # full_ds_dy = abs(float(full_ds_y_coords[1] - full_ds_y_coords[0]))
+        #
+        # num_extra_pixels = tile_enlargement
+        # extra_dx = (num_extra_pixels + 0.5) * full_ds_dx
+        # extra_dy = (num_extra_pixels + 0.5) * full_ds_dy
```
TonioF: If you have code that is commented out, you should always add a comment why it is. In this case I vote for removing it entirely.

TonioF (changes requested):

xcube/core/tile.py
clarasb: Yes, I agree. I remove it.

TonioF (approved):

forman (approved):

PR #1226, opened 2026-06-30
Remove submodule zarrstore

konstntokas:
Removed most Zarr store implementations from `xcube.core.zarrstore`, retaining only  `GenericZarrStore`. `GenericZarrStore` provides a flat `MutableMapping` view of a Zarr store, making it independent of any specific Zarr store implementation. This change prepares xcube for the migration to Zarr v3, whose store API differs significantly from that of Zarr v2.

Checklist:

* [x] GitHub CI passes


CHANGES.md
```diff
@@ -6,6 +6,11 @@
 
 ### Other changes
 * Constrained `matplotlib >=3.8.3,<3.11.0` (#1219)
+* Removed most Zarr store implementations from `xcube.core.zarrstore`, retaining only
+  `GenericZarrStore`. `GenericZarrStore` provides a flat `MutableMapping` view of a 
+  Zarr store, making it independent of any specific Zarr store implementation. 
+  This change prepares xcube for the migration to Zarr v3, whose store API differs 
+  significantly from that of Zarr v2.
```
forman, line 8: Since this is a breaking, non-minor API change, this should go into v1.14 if not even v2.0.

forman (changes requested):
LGTM, but since this is a API-breaking change we have two choices 

- bump the current dev version to 1.14.0.dev0 or
- create a new branch `releases/1.14` (better `releases/2.0`) and re-target this PR.  

Also address coverage drop.

CHANGES.md
konstntokas: @forman I thought xcube v2.0 is reserved for the decomposition of xcube. So I would go for v1.14. What do you think?

forman: It is just not nice to introduce a breaking change without prior deprecation.
But it is ok in this case, because I believe this concerns rarely used API.

forman (approved):
Ok, 1.14 it is.

Issue #1219, opened 2026-06-17
xcube incompatibilty with matplotlib 3.11.0 for custom colormaps


clarasb:
**Describe the bug**
xcube.util.cmaps.load_custom_colormap throws Value Error for categorical colormaps when using the latest version of `matplotlib`.

**To Reproduce**
Run e.g. the test `test.util.test_cmaps.ColormapRegistryTest.test_get_cmap_from_code_type_bound` or `xcube serve` with categorical custom colormap.

**Expected behavior**
The test runs without failure.

pont-us:
Hmm, [quite a lot of colourmap changes](https://matplotlib.org/stable/release/prev_whats_new/whats_new_3.11.0.html#colors-and-colormaps) in matplotlib 3.11, but on first glance I don't see anything that should break backward compatibility…

PR #1228, opened 2026-07-02
ensure compatibility with matplotblib 3.11.0

TonioF:
Solves #1219. The issue is that since matplotlib 3.11.0 color map values need to be given in a strictly ascending order; in our declaration of stepwise categorical color maps we state values as end points for one range and start points for the next one. I added a small value to the beginning of a range, so there is a tiny increase.

Checklist:

* ~[ ] Add unit tests and/or doctests in docstrings~
* [x] GitHub CI passes

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,7 +3,7 @@
 ### Fixes
 * Improved raster tile alignment by using nearest-pixel rounding during tile reprojection,
   reducing visual offsets between rendered map tiles and dataset coordinates. (#1216)
-* ensure compatibility with matplotlib 3.11.0 (#1219)
+* Ensure compatibility with matplotlib 3.11.0 (#1219)
 
 ### Other changes
 * Constrained `matplotlib >=3.8.3,<3.11.0` (#1219)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,7 +6,6 @@
 * Ensure compatibility with matplotlib 3.11.0 (#1219)
 
 ### Other changes
-* Constrained `matplotlib >=3.8.3,<3.11.0` (#1219)
 * Pinned libjxl <=0.11.2 because libjxl >=0.12.0 causes CI failures due to 
   rasterio/GDAL binary incompatibilities.
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 ### Fixes
 * Improved raster tile alignment by using nearest-pixel rounding during tile reprojection,
   reducing visual offsets between rendered map tiles and dataset coordinates. (#1216)
+* ensure compatibility with matplotlib 3.11.0 (#1219)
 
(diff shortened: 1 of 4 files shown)


clarasb (approved):
LGTM, but please do not forget to update `pyproject.toml` as well -> https://github.com/xcube-dev/xcube/blob/a62a48b18c59d1568e3c8d6838a3ce0be6071b3c/pyproject.toml#L39

CHANGES.md
```diff
@@ -3,6 +3,7 @@
 ### Fixes
 * Improved raster tile alignment by using nearest-pixel rounding during tile reprojection,
   reducing visual offsets between rendered map tiles and dataset coordinates. (#1216)
+* ensure compatibility with matplotlib 3.11.0 (#1219)
```
```suggestion
* Ensure compatibility with matplotlib 3.11.0 (#1219)
```

xcube/util/cmaps.py
```diff
@@ -522,10 +522,12 @@ def parse_cm_code(cm_code: str) -> tuple[str, Optional[Colormap]]:
             if cm_type == "stepwise":
                 # Turn cm_items into discrete step function
                 stepwise_cm_items = []
+                offset = 0
                 for i, (value, color) in enumerate(cm_items[0:-1]):
                     next_value = cm_items[i + 1][0]
-                    stepwise_cm_items.append((value, color))
+                    stepwise_cm_items.append((value + offset, color))
                     stepwise_cm_items.append((next_value, color))
+                    offset = 1e-8
```
forman, line 532: Please explain 

- the magic constant used here,
- why this does "ensure compatibility with matplotlib 3.11.0".

Comment in code.

environment.yml
```diff
@@ -23,7 +23,7 @@ dependencies:
   - jsonschema >=3.2
   - libgdal-jp2openjpeg
   - mashumaro
-  - matplotlib-base >=3.8.3,<3.11.0 # see Issue #1219
+  - matplotlib-base >=3.8.3
```
forman:
```suggestion
  - matplotlib-base >=3.8.3,<4
```

forman (changes requested):

xcube/util/cmaps.py
TonioF: The offset here stands for a very small value, just large enough to serve to create a difference in floating point numbers. The reason to set it is that when we create a series of stepwise_cm_items, matplotlib since v3.11.0 requires these to be strictly mononically increasing, otherwise an error is thrown and we continue with the fallback code.
I guess it makes sense to call the constant EPSILON to make it clearer.

Another option would have been to try to reimplement the former matplotlib code in xcube, but I preferred this solution.

forman: Just leave a comment in the code so someone else understands the magic.

forman (approved):

PR #1230, opened 2026-07-08
Add validation of query parameter for server requests

### LINKED ISSUE #1225: Review handling of unknown query parameters on `/api/viewer/`
During analysis of access logs, automated vulnerability scanners were observed sending requests such as:

```
GET /api/viewer/?rest_route=/wp/v2/users/
```

The request received a `200 OK` response although the query parameter is unrelated to xcube functionality.
It should be evaluated whether unknown query parameters should return `400` or `404`.

clarasb:
## Summary

This PR adds validation of allowed query parameters for requests to static routes and configures the `/api/viewer` route to reject unknown URL query parameters.
This prevents unrelated scanner requests such as `?rest_route=/wp/v2/users/` from being served as the Viewer index.html with `200 OK`.

## Testing

Run the server locally:

```powershell
xcube --loglevel INFO serve -c examples\serve\demo\config.yml  
```

In other terminal run:
```powershell
curl.exe http://localhost:8080/viewer/?rest_route=/wp/v2/users/ #Windows
curl http://localhost:8080/viewer/?rest_route=/wp/v2/users/ #Linux/macOS
```

Closes #1225.

Checklist:

* [x] GitHub CI passes

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,9 @@
   reducing visual offsets between rendered map tiles and dataset coordinates. (#1216)
 * Ensure compatibility with matplotlib 3.11.0 (#1219)
 * Fixed duplicated `tornado` log entries emitted by xcube Server. (#1224)
-
+* Added validation of allowed query parameters for static routes and 
+  configured allowed query parameters for `api/viewer/`. (#1225)
+  
 ### Other changes
 * Pinned libjxl <=0.11.2 because libjxl >=0.12.0 causes CI failures due to 
   rasterio/GDAL binary incompatibilities.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,7 +6,7 @@
 * Ensure compatibility with matplotlib 3.11.0 (#1219)
 * Fixed duplicated `tornado` log entries emitted by xcube Server. (#1224)
 * Added validation of allowed query parameters for static routes and 
-  configured allowed query parameters for `api/viewer/`. (#1225)
+  configured allowed query parameters for route `/viewer`. (#1225)
   
 ### Other changes
 * Pinned libjxl <=0.11.2 because libjxl >=0.12.0 causes CI failures due to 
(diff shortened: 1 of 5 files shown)


CHANGES.md
```diff
@@ -5,7 +5,9 @@
   reducing visual offsets between rendered map tiles and dataset coordinates. (#1216)
 * Ensure compatibility with matplotlib 3.11.0 (#1219)
 * Fixed duplicated `tornado` log entries emitted by xcube Server. (#1224)
-
+* Added validation of allowed query parameters for static routes and 
+  configured allowed query parameters for `api/viewer/`. (#1225)
```
```suggestion
  configured allowed query parameters for route `/viewer`. (#1225)
```

xcube/server/webservers/tornado.py
```diff
@@ -42,6 +42,45 @@
 
 SERVER_CTX_ATTR_NAME = "__xcube_server_ctx"
 
+def _assert_allowed_query_params(
+    request: tornado.httputil.HTTPServerRequest, allowed_query_params: Sequence[str]
+):
+    unknown_query_params = set(request.query_arguments) - set(allowed_query_params)
```
forman, line 48: Query parameters should be, afaik, not case-sensitive.

xcube/server/api.py
```diff
@@ -198,6 +202,9 @@ def static_route(
             path: The route path.
             default_filename: Optional default filename, e.g.,
                 "index.html".
+            allowed_query_params: Optional names of query parameters allowed
+                for this static route. If given, other query parameters should
+                be rejected by web framework implementations.
```
forman, line 207: So we only protect the routes that use this argument.

This is currently only `/viewer`. An routes of the form `/viewer/.../...`? What others routes like `/datasets`? 
Suggestion: `allowed_query_params` can also be a `bool`. If `True`, it can have any query params (like it is now). if `False`, any query parameters are forbidden. The new default would `False` and we need to check, which other endpoints must be protected.

forman (changes requested):
Looks good to me, but please consider my suggestions. We may have to discuss this further in more detail.

PR #1234, opened 2026-07-16
Fix tile rendering

clarasb:
This PR is a follow-up to #1216.

#1216 improved the tile indexing step by using nearest-pixel rounding when mapping tile pixels to dataset pixels. This fixed the actual tile shift, because target pixels are now sampled from the nearest source pixel instead of being consistently biased toward one side.

However, the source subset used for rendering was still computed too tightly. The subset bbox describes the transformed positions of the target tile pixel centers, but nearest-pixel rounding may select a source pixel whose center lies just outside that bbox.

```text
source pixel centers:
      x        x        x        x
      |--------|--------|--------|

transformed target bbox:
           |----------------|
           ^                ^
   nearest source pixel     
   may lie outside bbox     
```
If such an edge source pixel was not included in the loaded subset, it was treated as invalid and rendered as missing or transparent data. **This PR fixes** this behaviour by enlarging the source subset now using the dataset’s own pixel size, including the half-pixel margin required by nearest-pixel rounding.

Example: `xcube serve -vvv -c examples/serve/panels-demo/config.yaml`

image

Checklist:

* [x] GitHub CI passes

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,7 @@
  
 ### Fixes
 * Improved raster tile alignment by using nearest-pixel rounding during tile reprojection,
-  reducing visual offsets between rendered map tiles and dataset coordinates. (#1216)
+  reducing visual offsets between rendered map tiles and dataset coordinates. (#1216, #1234)
 * Ensure compatibility with matplotlib 3.11.0 (#1219)
 * Fixed duplicated `tornado` log entries emitted by xcube Server. (#1224)
 
--- a/xcube/core/tile.py
+++ b/xcube/core/tile.py
@@ -138,6 +138,7 @@ def compute_tiles(
     with measure_time("Transforming tile map to dataset coordinates"):
         ds_x_name, ds_y_name = ml_dataset.grid_mapping.xy_dim_names
 
+        ds_x_coords = variable_0[ds_x_name]
         ds_y_coords = variable_0[ds_y_name]
         ds_y_points_up = bool(ds_y_coords[0] < ds_y_coords[-1])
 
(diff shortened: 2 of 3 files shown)


CHANGES.md
```diff
@@ -2,7 +2,7 @@
  
 ### Fixes
 * Improved raster tile alignment by using nearest-pixel rounding during tile reprojection,
-  reducing visual offsets between rendered map tiles and dataset coordinates. (#1216)
+  reducing visual offsets between rendered map tiles and dataset coordinates. (#1216, #1234)
```
TonioF, line 5:
```suggestion
  Reducing visual offsets between rendered map tiles and dataset coordinates. (#1216, #1234)
```

TonioF (commented):
Just one comment, otherwise it looks good to me.

xcube/core/tile.py
```diff
@@ -138,6 +138,7 @@ def compute_tiles(
     with measure_time("Transforming tile map to dataset coordinates"):
         ds_x_name, ds_y_name = ml_dataset.grid_mapping.xy_dim_names
 
+        ds_x_coords = variable_0[ds_x_name]
         ds_y_coords = variable_0[ds_y_name]
```
forman, line 142: If not already asserted upstream, consider:

```suggestion
        ds_x_coords = variable_0[ds_x_name]
        ds_y_coords = variable_0[ds_y_name]
        assert ds_x_coords.size > 1
        assert ds_y_coords.size > 1
```

forman (approved):
Good, thanks!

Pls, see my comment!

xcube/core/tile.py
clarasb: Thanks for the suggestion, but not needed as assertion already happens upstream.

PR #1238, opened 2026-07-26
Use pixi development environment

### LINKED ISSUE #1237: Migrate xcube development environments to Pixi
## Summary

Replace the current `environment.yml`-based development environments with Pixi across the xcube ecosystem.

The goal is to simplify environment setup, provide reproducible dependency resolution via lock files, and modernize the developer workflow while keeping the existing repository structure unchanged.

We should keep `rtd-environment.yml` since RTD specifically relies on it today. But the long-term plan is be to replace RTD anyway.
Should this take longer, we may define a pixi _feature_ `"docs"`.

## Motivation

Today, each xcube repository provides an `environment.yml` that developers create using Conda or Mamba.

Pixi provides several advantages:

- reproducible environments through `pixi.lock`
- significantly simpler onboarding (`pixi install`)
- integrated task runner
- unified Conda + PyPI dependency management
- cross-platform lock files
- no need for developers to manually manage Conda environments

## Proposed approach

Each repository should become an independent Pixi project. However, as a first step, only **xcube** should be migrated, plugins and companion libraries should be migrated separately. See #1239.

Instead of

```
environment.yml
```

each repository would contain

```
pyproject.toml
pixi.lock
```

with the Pixi configuration stored in `pyproject.toml` under `tool.pixi`.

The existing `pyproject.toml` already contains the package metadata, so no additional `pixi.toml` should be required:

```bash
pixi init --format pyproject
pixi import environment.yml
pixi install
pixi run pytest
git add pixi.lock
git remove environment.yml
```

Example repository layout:

```
xcube/
    pyproject.toml
    pixi.lock

xcube-stac/
    pyproject.toml
    pixi.lock

xcube-cmems/
    pyproject.toml
    pixi.lock
```

## Developer workflow

Current:

```bash
mamba env create -f environment.yml
conda activate ...
pip install -e .
pytest
```

Proposed:

```bash
pixi install
pixi run tests
```

or

```bash
pixi shell
```

followed by the usual development commands.

## Development model

This proposal does **not** suggest moving to a monorepo or a shared Pixi workspace.

Each repository remains independently developable and maintains its own lock file.

This means plugin development changes very little compared to today.

## Developing xcube together with a plugin

Plugin repositories should normally depend on a released xcube version.

When simultaneously developing xcube and a plugin, developers should be able to temporarily use a local editable checkout of xcube (e.g. via a local path dependency) without committing those local paths to the repository.

```toml
[tool.pixi.pypi-dependencies]
xcube = { path = "../xcube", editable = true }
xcube-cmems = { path = ".", editable = true }
```

The exact workflow should be documented, but this proposal intentionally keeps local cross-repository development separate from the default development setup.

## Migration plan

- migrate each `environment.yml` to Pixi
- store Pixi configuration in `pyproject.toml`
- commit `pixi.lock`
- remove `environment.yml`
- update contributor documentation
- optionally define common Pixi tasks (`test`, `lint`, `format`, `docs`, etc.)

## Benefits

- simpler onboarding
- reproducible environments
- faster environment creation
- integrated task runner
- one project configuration file (`pyproject.toml`)
- consistent developer experience across all xcube repositories
- no change to the existing multi-repository development model

## Recommended pixi task set

The following definitions allow for conveniently running the common, must-apply tasks for xcube and xcube plugins

- `pixi run format`  _run all code formatters_ 
- `pixi run checks`  _run all code checkers_ 
- `pixi run tests`  _run all tests_ 

and also

- `pixi run jl`  _run jupyter-lab_ 
- `pixi run doc-serve`  _run mkdocs serve_
- `pixi run doc-build`  _run mkdocs build_

Copy the following code block to the end of the `pyproject.toml` files of xcube and xcube plugins:
 
```toml
# pixi run ...

[tool.pixi.tasks]
# Top-level tasks
doc-serve = "mkdocs serve"
doc-build = "mkdocs build"
jl = "jupyter-lab"
tests = "pytest tests"
# The following are helpers for composite tasks that use `depends-on`
format-with-isort = "isort src tests"
format-with-ruff = "ruff format src tests"
check-with-ruff = "ruff check src"
check-with-mypy = "mypy src"
cov-base = "pytest --cov src/sen4cap_client --cov-report= --cov-append tests"
cov-report-html = "coverage html -d .cov-report && coverage report"
cov-report-xml = "coverage xml -o coverage.xml && coverage report"

# pixi run format

[tool.pixi.tasks.format]
depends-on = ["format-with-isort", "format-with-ruff"]

# pixi run checks

[tool.pixi.tasks.checks]
depends-on = ["check-with-ruff", "check-with-mypy"]

# pixi run coverage

[tool.pixi.tasks.coverage]
depends-on = ["cov-base", "cov-report-html"]

# pixi run coverage-ci

[tool.pixi.tasks.coverage-ci]
depends-on = ["cov-base", "cov-report-xml"]
```

forman:
Closes #1237 

This PR does not yet delete our `environment.yml` as this would break current CI. 
This should be done in a subsequent PR.

Checklist:

* [x] GitHub CI passes


forman:
FYI 

- @b-yogesh 
- @clarasb 
- @freds-dev 
- @konstntokas 
- @TejasMorbagal 
- @thomasstorm 
- @TonioF

pyproject.toml
```diff
@@ -17,7 +14,7 @@ keywords = [
   "datacube", "xarray", "zarr"
 ]
 license = {text = "MIT"}
-requires-python = ">=3.9"
+requires-python = ">=3.10,<3.14"
 dependencies = [
```
pont-us, line 17:
```suggestion
dependencies = [
"affine>=2.2",
"altair",
```

pyproject.toml
pont-us:
```suggestion
  "click>=8.2.0",
```

pont-us:
```suggestion
  "dask>=2024.8",
```

pont-us:
```suggestion
  "tornado>=6.0,<6.5",
```

pont-us:
```suggestion
  "urllib3>=2.0",
```

pont-us, line 100:
```suggestion
```

Assuming that the removal of `version` from `dynamic` above was intentional, this assignment needs to be removed as well.

pont-us (changes requested):
There were some mismatches between the package versions specified in `project.dependencies` and those in `tool.pixi.dependencies`.  I've assumed that the latter is the source of truth and suggested the appropriate changes for the former.

`affine>=2.2` and `altair` are missing from `project.dependencies`. I added them both. `affine` is used by `xcube.core.resampling` so clearly needed adding; `altair` is only needed for `chartlets` but I assume that we want to keep the two dependency sets in sync so I added it too.

(`libgdal-jp2openjpeg` and `libjxl` are also only present in
`tool.pixi.dependencies` but of course this is expected since they're not
available on PyPI.)

pyproject.toml
```diff
@@ -1,10 +1,7 @@
-[build-system]
-requires = ["setuptools >= 61.2.0"]
-build-backend = "setuptools.build_meta"
-
 [project]
-name = "xcube" # PYPI RENAME THIS LINE (do not remove this comment, see #1010) 
-dynamic = ["version", "readme"]
+name = "xcube" # PYPI RENAME THIS LINE (do not remove this comment, see #1010)
+version = "1.13.4.dev0"
```
konstntokas: But isn't it better if we keep it dynamical?

pyproject.toml
konstntokas: https://github.com/xcube-dev/xcube/pull/1238/changes#r3691331891 needs to be resolved first.

pyproject.toml
konstntokas: If we do this we have two sources of truth, in `version.py` and in `pyproject.toml`.

pont-us: Good point. On reflection, I think that this change must have leaked in from PR #1242 , where it makes sense (because that PR is precisely about moving the SSOT from `version.py` to `pyproject.toml`, and includes corresponding changes to `version.py` to avoid duplication). So I agree we should retain `dynamic = ["version", "readme"]` in this PR, ignore my related suggestion in `[tool.setuptools.dynamic]`, and implement these changes in #1242 instead.

konstntokas: I will revert these changes then and we do them in #1242

pyproject.toml
konstntokas: We keep it here and do the change in #1242

konstntokas (approved):

pont-us (approved):

forman:
> There were some mismatches between the package versions specified in `project.dependencies` and those in `tool.pixi.dependencies`. I've assumed that the latter is the source of truth and suggested the appropriate changes for the former.

The other way round.  The `project.dependencies` are the true package/library dependencies and are used for building the package. `tool.pixi.dependencies` and `tool.pixi.pypi-dependencies` are the dependencies for the development environment of the project. Other environment configurations may be added.

PR #1240, opened 2026-07-26
Optimize multi-level dataset loading of non-spatial coordinates

### LINKED ISSUE #1236: Optimization: reuse non-spatial dimension arrays in multi-level datasets
Handling of non-spatial dimensions for MLDatasets can be optimized.
Non-spatial dimension arrays should stay the same for all dataset levels and actually could be reused from level zero.

**Suggested enhancements:**

The non-spatial dimension values should be read only one time, for level 0. 
For all other levels non-spatial dims should be dropped when opening the level and assigned afterwards.

```
ds_0 = xr.open_dataset("data/level_0.zarr")

ds_1 = xr.open_dataset("data/level_1.zarr", drop_variables=["time"])
ds_1 = ds.assign_coords(time=ds_0.time)
```

forman:
Closes #1236

## Summary

Optimizes opening multi-level datasets by reusing invariant, non-spatial coordinates from level zero instead of reading them again for every level.

Coordinates that use spatial dimensions, including `lat_bnds` and `lon_bnds`, remain level-specific and are excluded from reuse.

## Changes

- Add a shared helper for opening Zarr-backed dataset levels.
- Identify reusable coordinates based on their dimensions.
- Reuse non-spatial coordinates such as `time` and `time_bnds`.
- Preserve spatial coordinates and bounds from each individual level.
- Add regression tests for coordinate selection and level opening.

## Testing

- `test/core/mldataset/test_fs.py`: 7 tests passed.
- Ruff checks passed.
- `git diff --check` passed.

Checklist:

* [x] GitHub CI passes


forman:
FYI @clarasb

forman:
FYI @martin-boettcher

CHANGES.md
```diff
@@ -1,55 +1,56 @@
...
-  Zarr store, making it independent of any specific Zarr store implementation. 
-  This change prepares xcube for the migration to Zarr v3, whose store API differs 
-  significantly from that of Zarr v2.
-* Migrated to Zarr v3, adapting xcube's Zarr handling to the new Zarr store API and
-  ensuring compatibility with the latest Zarr ecosystem.
-
-## Changes in 1.13.4
+## Changes in 1.13.4 (in development)
  
+### Enhancements
```
konstntokas, line 14: I think this should go under 

```
## Changes in 1.14.0 (in development)
```

xcube/core/mldataset/fs.py
```diff
@@ -370,14 +362,71 @@ def write_dataset(
...
+                        index=index,
+                        consolidated=consolidated,
                     )
-                    level_dataset.zarr_store.set(level_zarr_store)
                     ml_dataset.set_dataset(index, level_dataset)
 
         return path
 
+    @classmethod
+    def _open_level_dataset(
```
konstntokas, line 385: I expected that the non-spatial coordinates are cached in the class when initiating it. I do not understand, how this can optimize the access. For each layer, instead of taking the coordinates from the requested layer, it now takes it from the base layer.

konstntokas: When writing to `.levels`, the non-spatial coordinates are still written for each layer, right?

konstntokas (commented):
The changes looks fine, but I do not understand, how this should optimize the opening. Furthermore, branch cannot be merged due to conflicts.

xcube/core/mldataset/fs.py
forman: > I expected that the non-spatial coordinates are cached in the class when initiating it. I do not understand, how this can optimize the access. For each layer, instead of taking the coordinates from the requested layer, it now takes it from the base layer.

Since the non-spatial coordinates should be the same for each level, we do no longer read them from source for all levels > 0, as this can be slow (e.g., chunked coordinate arrays on remote S3). Instead we simply reuse the level 0 coordinates.

forman: > When writing to `.levels`, the non-spatial coordinates are still written for each layer, right?

Yes, because each level should be a self-contained zarr dataset.

PR #1242, opened 2026-07-27
Get version from `pyproject.toml`

forman:
The xcube version identifier is now retrieved from `pyproject.toml` as single source of truth. 

Checklist:

* [x] GitHub CI passes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,10 +1,7 @@
-[build-system]
-requires = ["setuptools >= 61.2.0"]
-build-backend = "setuptools.build_meta"
-
 [project]
-name = "xcube" # PYPI RENAME THIS LINE (do not remove this comment, see #1010) 
-dynamic = ["version", "readme"]
+name = "xcube" # PYPI RENAME THIS LINE (do not remove this comment, see #1010)
+version = "1.14.0.dev0"
+dynamic = ["readme"]
 authors = [
   {name = "xcube Development Team"}
 ]
@@ -85,6 +82,10 @@ xcube = "xcube.cli.main:main"
 # entry point xcube's default extensions
 [project.entry-points.xcube_plugins]
 xcube = "xcube.plugin:init_plugin"
+
+[build-system]
+requires = ["setuptools >= 61.2.0"]
+build-backend = "setuptools.build_meta"
 
 [tool.setuptools.package-data]
 "xcube.webapi.meta" = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -96,7 +96,6 @@ build-backend = "setuptools.build_meta"
 ]
 
 [tool.setuptools.dynamic]
-version = {attr = "xcube.__version__"}
 readme = {file = "README.md", content-type = "text/markdown"}
 
 [tool.setuptools.packages.find]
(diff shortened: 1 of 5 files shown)


readthedocs.yml
```diff
@@ -14,4 +14,9 @@ sphinx:
 conda:
   environment: rtd-environment.yml
 
+python:
+  install:
+    - method: pip
+      path: .
+
```
pont-us, line 21: Is this connected with the main purpose of the PR, or just thrown in as an incidental improvement? (As far as I know these are the defaults anyway, but I don't see any harm in making them explicit.)

pyproject.toml
pont-us:
```suggestion
```

If it's no longer dynamic, we shouldn't be setting it here.

xcube/version.py
```diff
@@ -2,4 +2,11 @@
 # Permissions are hereby granted under the terms of the MIT License:
 # https://opensource.org/licenses/MIT.
 
-version = "1.14.0.dev0"
+from importlib.metadata import PackageNotFoundError, version as get_version
+
+try:
+    # xcube on conda-forge and editable pip installs
+    version = get_version("xcube")
```
pont-us, line 9: There's one case that's not covered here: if xcube is not installed as a distribution package, but is available in `sys.path` as an import package. The most common example is when the xcube repo root is the CWD. In that case, `import xcube` has worked as expected up to now, but on this PR branch it results in "importlib.metadata.PackageNotFoundError: No package metadata was found for xcube-core" because `importlib.metadata` only examines distribution packages. I think that we should continue to support this case as well, or at the very least catch it explicitly and give the user an informative error message.

pont-us (changes requested):
We need to decide how to handle the case where xcube is on the Python path, but not installed as a distribution package -- see comment on `version.py`.

xcube/version.py
```diff
@@ -2,4 +2,11 @@
...
+except PackageNotFoundError:
+    # On PyPI, xcube is called xcube-core
+    version = get_version("xcube-core")
```
pont-us:
```suggestion
    try:
        # On PyPI, xcube is called xcube-core
        version = get_version("xcube-core")
    except PackageNotFoundError:
        # If neither distribution package is installed, look for the
        # project file relative to this source file.
        from pathlib import Path
        import tomllib
        with open(Path(__file__).parent.parent / "pyproject.toml", "rb") as f:
            version = tomllib.load(f)["project"]["version"]```

readthedocs.yml
konstntokas: I think, this was needed because xcube's version can be only derived when installed, as described in the comment https://github.com/xcube-dev/xcube/pull/1242/changes#r3681622297. With the suggestions https://github.com/xcube-dev/xcube/pull/1242/changes#r3691069843 I would expect that it works again without these lines.

konstntokas: Nevertheless I would just keep them as they are not harmful and common practice.

xcube/version.py
konstntokas: suggestions applied.

konstntokas (approved):

readthedocs.yml
pont-us: Agreed. Explicit is better than implicit :)

pont-us (approved):

